PR #24, opened 2022-06-10
feat: devenv.yaml

george-e-shaw-iv:

* adds a devenv.yaml file to track various things a devenv may need to
  deploy a service. right now we're only tracking dependencies.
* this allows other stencil modules to programmatically add depedencies
  via module hooks.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/devenv.yaml.tpl
+++ b/templates/devenv.yaml.tpl
@@ -1,0 +1,17 @@
+dependencies:
+{{- if gt (len (stencil.GetModuleHook "devenv.dependencies.optional")) 0 }}
+	{{- range (stencil.GetModuleHook "devenv.dependencies.optional") }}
+	optional:
+		- "{{ . }}"
+	{{- end }}
+{{- else }}
+	optional: []
+{{- end }}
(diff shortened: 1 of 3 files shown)


plisy (approved):

templates/devenv.yaml.tpl
```diff
@@ -0,0 +1,17 @@
+dependencies:
+{{- if gt (len (stencil.GetModuleHook "devenv.dependencies.optional")) 0 }}
+	{{- range (stencil.GetModuleHook "devenv.dependencies.optional") }}
+	optional:
+		- "{{ . }}"
+	{{- end }}
+{{- else }}
+	optional: []
+{{- end }}
+{{- if gt (len (stencil.GetModuleHook "devenv.dependencies.required")) 0 }}
```
jaredallard, line 10: Unncessary if, range will noop if no values

jaredallard (approved):

templates/devenv.yaml.tpl
george-e-shaw-iv: i need to put an empty array in the yaml file if it has no values

PR #30, opened 2022-06-21
feat: add monitoring terraform templates

artemchernyak-outreach:

Migrates the monitoring terraform templates from bootstrap. This is needed before we add OpenTelemetry support to the automatically generated alerts and dashboards.

## Jira ID

[DTSS-1911]

## Notes for your reviewers

[DTSS-1911]: https://outreach-io.atlassian.net/browse/DTSS-1911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,5 +1,7 @@
 name: github.com/getoutreach/stencil-golang
 ###Block(keys)
+modules:
+  - name: github.com/getoutreach/stencil-discovery
 arguments:
   service:
     type: bool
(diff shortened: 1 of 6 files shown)


bhogatemadhura:
as discussed in internal channel, can we only generate these terraforms if the service is getting deployed ?

artemchernyak-outreach:
@bhogatemadhura I think we should do that change in a separate PR. That way we can modify and back it out separately.

bhogatemadhura:
@artemchernyak-outreach That’s fair. May be we can create a ticket for that separately

bhogatemadhura (approved):

PR #34, opened 2022-07-05
fix(DTSS-2002): Correct handling of hyphens in service names

lelandbatey:

This change fixes broken codegen when the `name` in service.yaml contains hyphen (`-`) character(s). Hyphens are allowed in service names, but the names of Services in Protobuf and the names of structs in Go are *not* syntactically allowed to have hyphens.

Till now, these stencil templates assumed that the service name would be a valid identifier in all the languages that these templates generate code for. This would lead to Stencil generating code which was invalid and would not compile.

This new code handles these cases with the following text transformations to the service name (only where situationally required):

1. Replace all non-word characters with spaces
2. "Title case" the words in the string (capitalize the first letter of each group of characters in each series of letters separated by space characters).
3. Remove the spaces in the newly title-cased series of words

The effect of this set of transformations is as such:

```
"foo-bar"             becomes "FooBar"
"Fizz-Bang"           becomes "FizzBang"
"ZIP-WOW"             becomes "ZIPWOW"
"plop-crunch-splat"   becomes "PlopCrunchSplat"
"thud_pow-baz"        becomes "ThudPowBaz"
"click.clack|blob"    becomes "ClickClackBlob"
```

This approach is certainly not bullet-proof, but seems reasonable for the moment, and fixes code generation for a number of existing services. Here are some of the services which have their codegen fixed by this change (if they were ever to become gRPC services):

```
github.com/getoutreach/cli-service
github.com/getoutreach/ds-goldtables
github.com/getoutreach/ectoken-go
github.com/getoutreach/fluffy-parakeet
github.com/getoutreach/go-mailroom
github.com/getoutreach/go-outreach
github.com/getoutreach/kaia-slack
github.com/getoutreach/mailroom-backfill-producer
github.com/getoutreach/mailroom-indexer
github.com/getoutreach/meeting-verifier
github.com/getoutreach/mysql-bootstrap-flagship-bento
github.com/getoutreach/outreach-accounts
github.com/getoutreach/outreach-airflow-ecs
github.com/getoutreach/outreach-vault-database-plugin
github.com/getoutreach/proxysql-hostgroup-refresher
github.com/getoutreach/shard-manager-data-agent
```

## Jira IDs

https://outreach-io.atlassian.net/browse/DTSS-2002 [DTSS-2002]
https://outreach-io.atlassian.net/browse/ORGS-2236 [ORGS-2236]

## Notes for reviewers

This change doesn't include tests for this corrected behavior, as I could not find where such tests would go. Please let me know where I should place tests for verifying we're generating valid code, if there is such a place.

[DTSS-2002]: https://outreach-io.atlassian.net/browse/DTSS-2002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/api/api.go.tpl
```diff
@@ -14,7 +14,7 @@ import (
 	"context"
 )
 
-// Service is the {{ .Config.Name | title }} server interface
+// Service is the {{ regexReplaceAllLiteral "\\W+" .Config.Name " " | title | replace " " "" }} server interface
```
artemchernyak-outreach, line 17: Since you use this in a few places, and it might be used in more in the future, you can drop this into https://github.com/getoutreach/stencil-golang/blob/main/templates/_helpers.tpl and reference it here.

artemchernyak-outreach (approved):

templates/api/api.go.tpl
lelandbatey: Excellent, I'll move it to that location.

lelandbatey: These have been move to the `_helpers.tpl` file.

templates/main_test.go
```diff
@@ -0,0 +1,25 @@
...
+var requiredtmpls = []string{
+	"_helpers.tpl",
+}
+
+func TestRenderAPIGoSuccess(t *testing.T) {
+	// NOTE: 2022-07-06 For the moment, we cannot change the `Name` field of
+	// the ServiceManifest used by the `Run()` method in stenciltest, which is
+	// why this test does not verify correct handling of odd service names.
+	st := stenciltest.New(t, "api/api.go.tpl", requiredtmpls...)
+	st.Run(false)
```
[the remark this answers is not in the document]
ChrisPRobinson: Looks good to me.

ChrisPRobinson (approved):
LGTM

PR #38, opened 2022-07-13
feat: add e2e profile to devspace

artemchernyak-outreach:

Adds e2e profile so we can run e2e tests inside devspacefd

## Jira ID

[DTSS-1601]

## Notes for your reviewers

[DTSS-1601]: https://outreach-io.atlassian.net/browse/DTSS-1601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/devspace.yaml.tpl
+++ b/templates/devspace.yaml.tpl
@@ -310,8 +310,15 @@ profiles:
       - op: add
         path: hooks
         value:
-          name: reset-dev
+          name: reset-dev-interrupt
           events: ["devCommand:interrupt"]
+          command: |-
+            "${DEVENV_DEVSPACE_BIN}" reset pods -s
+      - op: add
+        path: hooks
+        value:
+          name: reset-dev-error
+          events: ["error:sync:app"]
           command: |-
             "${DEVENV_DEVSPACE_BIN}" reset pods -s
       - op: add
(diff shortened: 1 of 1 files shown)


george-e-shaw-iv (approved):

templates/devspace.yaml.tpl
```diff
@@ -310,10 +310,17 @@ profiles:
...
-          name: reset-dev
+          name: reset-dev-interrupt
           events: ["devCommand:interrupt"]
           command: |-
             "${DEVENV_DEVSPACE_BIN}" reset pods -s
+      - op: add
+        path: hooks
+        value:
+          name: reset-dev-error
+          events: ["error:sync:app"]
```
bhogatemadhura, line 321: when does this event occur ?

bhogatemadhura (approved):

templates/devspace.yaml.tpl
artemchernyak-outreach: This event occurs when the e2e test fails. I want to make sure that we clean everything up so that it's a fresh run next time. 

Kuberenets throws errors about multiple pods using mounting the same secrets if we don't clean it up properly.

PR #47, opened 2022-07-27
fix(monitoring): generate only if a service

jaredallard:

Title says, it does. I don't think we meant to generate these for non-service repositories.

## Jira ID

[XX-XX]


george-e-shaw-iv (approved):
GOOD CALL JARED

PR #62, opened 2022-08-11
feat(golang): 1.19

jaredallard:

This PR moves us to use `go 1.19` by default, and allows the go version to be (partially, if with bootstrap) overriden via the `versions.go` key.

## Jira ID

[XX-XX]

### DIFF
--- a/.releaserc.yaml
+++ b/.releaserc.yaml
@@ -1,6 +1,6 @@
 preset: conventionalcommits
 branches:
-  - name: rc
+  - name: main
     prerelease: "rc"
   - name: release
 plugins:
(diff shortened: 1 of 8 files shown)


.releaserc.yaml
```diff
@@ -1,6 +1,6 @@
 preset: conventionalcommits
 branches:
-  - name: rc
+  - name: main
```
artemchernyak-outreach, line 3: Did you re-stencil this? Not sure why this got changed, but we definitely don't want this.

jaredallard: I think an earlier version of `stencil-base` was used

jaredallard: I'll just re-run w/ frozen-lockfile

george-e-shaw-iv (approved):

PR #74, opened 2022-09-06
feat: support hyphenated service names for node clients

ikokli:
Currently hyphenated service names will cause issues when generating node clients.  The issue comes from the template using `title .Config.Name` which retains the hypens and is used in imports and function names.

The fix is to use already provided stencil helper `serviceNameLanguageSafe` on the node js templates.

## Jira ID

[DT-2791]

## Notes for your reviewers

First PR to the stencil repo, I am not sure how I can test or apply this.  Any guidance would be greatly appreciated.

[DT-2791]: https://outreach-io.atlassian.net/browse/DT-2791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/api/clients/node/src/client-helpers.ts.tpl
```diff
@@ -27,9 +27,9 @@ export interface ClientOptions {
 /**
  * @param accessToken The token to use to authenticate all requests the client submits
  * @param options The client options that affect its behavior
- * @returns The newly created {{ title .Config.Name }}Client instance
+ * @returns The newly created {{ stencil.ApplyTemplate "serviceNameLanguageSafe" }}Client instance
```
artemchernyak-outreach, line 30: Nit: you could avoid running this template everywhere by setting it to a variable.

e.g. add the below at. Close to the top of the file and reference it throughout.

```
{{- $serviceName := stencil.ApplyTemplate "serviceNameLanguageSafe" }}
```

artemchernyak-outreach (approved):

PR #79, opened 2022-09-09
feat(monitoring) specify tags to be added to metrics

kaldorn:
Limits the tags that are on Go metrics emitted by stencil services.

## Jira ID

Part of [DT-2728]

## Notes for your reviewers

[DT-2728]: https://outreach-io.atlassian.net/browse/DT-2728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


artemchernyak-outreach (approved):
This is awesome! Nice find on being able to configure this.

jaredallard (approved):

templates/monitoring/metric_tag_configuration.tf.tpl
```diff
@@ -0,0 +1,86 @@
...
+    "{{ .Config.Name }}.go_memstats_mspan_inuse_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_mspan_sys_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_next_gc_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_other_sys_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_stack_inuse_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_stack_sys_bytes": "gauge",
+    "{{ .Config.Name }}.go_memstats_sys_bytes": "gauge",
+    "{{ .Config.Name }}.go_sched_goroutines_goroutines": "gauge",
+    "{{ .Config.Name }}.go_sched_latencies_seconds": "distribution",
+    "{{ .Config.Name }}.go_threads": "gauge",
```
artemchernyak-outreach, line 65: It would be nice to add a block here and in the tags section. That way teams can expand this list as they see fit.

Maybe also a third block at the bottom of the file that would allow completely custom `datadog_metric_tag_configuartion` blocks.

templates/monitoring/metric_tag_configuration.tf.tpl
```diff
@@ -0,0 +1,86 @@
+{{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}
+## We use a locals block to have a single point of truth for the metric names, types, and tags that we want all metrics to have
+locals {
+  # Defines the metric names and types that we want to limit tags for
+  metrics = {
+    ## Go metrics
```
jkinkead, line 6: How much are these metrics actually being used? Picking a particularly large one (`go_gc_heap_allocs_by_size_bytes_total`), it seems like it isn't being used _at all_ in dashboards or anything.

These metrics all-told account for about 10% of our Datadog custom metric spend - so eliminating them could save us about 50k per year.

These all look like low-level information that's only going to be interesting in edge-case debugging. Can we just flag these off, and less customers enable them when they _really_ need them for live debugging?

templates/monitoring/metric_tag_configuration.tf.tpl
```diff
@@ -0,0 +1,86 @@
...
+    "{{ .Config.Name }}.go_threads": "gauge",
+  }
+
+  # Defines the tags to be added to the metrics above -- anything not on this list is not included
+  tags = [
+    "app",
+    "bento",
+    "cluster_name",
+    "env",
+    "reporting_team",
```
jkinkead, line 74: Are you sure this shouldn't be `"outreach_team"`? I don't see this tag in Datadog.

templates/monitoring/metric_tag_configuration.tf.tpl
```diff
@@ -0,0 +1,86 @@
...
+    "pod_name",
+    "region",
```
jkinkead, line 76: This tag set seems likely to reduce cardinality by 80-90%, which is great. Have you run a full analysis on the potential savings?

templates/monitoring/metric_tag_configuration.tf.tpl
kaldorn: Done

artemchernyak-outreach (approved):

jkinkead:
Closing for now. I think this is largely obsolete with the metric tag reducer.

It might be good to revive when Nate's back from leave.

PR #83, opened 2022-09-12
fix: do not render devenv.yaml for non-services

chatchai-outreach:

devenv.yaml should not render for non-services (libraries and CLIs)

## Jira ID

https://outreach-io.atlassian.net/browse/DT-2176

### DIFF
--- a/templates/devenv.yaml.tpl
+++ b/templates/devenv.yaml.tpl
@@ -1,3 +1,4 @@
+{{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}
 service: {{ stencil.Arg "service" }}
 dependencies:
 {{- if gt (len (stencil.GetModuleHook "devenv.dependencies.optional")) 0 }}


templates/devenv.yaml.tpl
```diff
@@ -1,3 +1,4 @@
+{{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}
```
jaredallard, line 1: I think we might've actually had a reason for this... do you recall @george-e-shaw-iv?

george-e-shaw-iv: We wanted to be able to allow libraries to run e2e tests

chatchai-outreach: so we no longer need https://outreach-io.atlassian.net/browse/DT-2176?

george-e-shaw-iv: Thats up for y'all to decide

chatchai-outreach:
per discussed with @jaredallard, we no longer it

PR #85, opened 2022-09-13
fix: upgrade protoc to 21.5, aka 3.21.5

nissim-natanov-outreach:
We have recently moved to protoc 3.21.5 (released under tag = 21.5). However, stencil-golang keeps reverting it back. Putting a patch here to ensure it sticks.

## Jira ID

[CDT-00]

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -23,7 +23,7 @@
   version: 2.6.6
 # Used in CI
 - name: protoc
-  version: 3.19.1
+  version: 21.5
 {{- end }}
 
 # Registers our versions w/ stencil-base


templates/_helpers.tpl
```diff
@@ -23,7 +23,7 @@
   version: 2.6.6
 # Used in CI
 - name: protoc
-  version: 3.19.1
+  version: 21.5
```
nissim-natanov-outreach, line 26: It is 21.5 (and not 3.21.5).
They changed the release tags from some reason while protoc itself keeps reporting 3.21.5 when you run `protoc --version`).

jaredallard:
You've tested this on a gRPC artifact generating repository, right? Just want to make sure it actually works, since I think we had to do the `3.` schema for asdf to work properly

nissim-natanov-outreach:
> You've tested this on a gRPC artifact generating repository, right? Just want to make sure it actually works, since I think we had to do the `3.` schema for asdf to work properly

yes, clerkcommons has protoc = 21.5 for a while now: https://github.com/getoutreach/clerkcommons/blob/main/.tool-versions#L6

nissim-natanov-outreach:
@jaredallard  you can also see the removal of 3. prefix from the release here: https://github.com/protocolbuffers/protobuf/tags. All latest releases have no 3. prefix - dunno why they did it.

jaredallard:
Weird, oddly enough there is still a `v3.21.5` tag there, but if it works... it works 🤷 

> @jaredallard you can also see the removal of 3. prefix from the release here: https://github.com/protocolbuffers/protobuf/tags. All latest releases have no 3. prefix - dunno why they did it.

jaredallard (approved):
Assuming there's no breaking changes in this protoc release, LGTM

nissim-natanov-outreach:
> `v3.21.5`

It does not work - the tag itself unfortunately does not have the binaries we need. Only the release has it. You can see diff here:
* https://github.com/protocolbuffers/protobuf/releases/tag/v3.21.5 - only source files
* https://github.com/protocolbuffers/protobuf/releases/tag/v21.5 - all the binaries

🤷

artemchernyak-outreach:
@nissim-natanov-outreach Do you need `21.5` specifically?

There is an issue with the js generation https://github.com/protocolbuffers/protobuf-javascript/issues/127. The only recommended solution is to downgrade to `3.20`.

You can see the error here: https://app.circleci.com/pipelines/github/getoutreach/services/1302/workflows/a29efb65-a50a-40cd-9da4-6da062e44a74/jobs/4694

PR #97, opened 2022-09-19
fix: bump orgservice

artemchernyak-outreach:

We need to ensure that teams update to the latest version so that the generate `orgservice` code works as expected.

## Jira ID

[XX-XX]


jaredallard (approved):
Doesn't do anything w/o the go mod processor, but LGTM to me for now. I'll revamp this with a module hook soon

PR #98, opened 2022-09-20
feat: generate/merge go.mod

jaredallard:

This PR migrates the original bootstrap `go.mod` processor to stencil, in a stencil native way (a extension!). This PR contains the new go module "processor" (extension) and the `go.mod` template code needed to use it.

This `go.mod` function merges the on disk `go.mod` with a generated one from dependencies from module hooks, and baked in dependencies. This is done by adding a extension with `MergeGoMod` that merges the right hand `go.mod` into the left hand one, keeping dependencies if they are higher than the right hand `go.mod`

## Jira ID

[DT-2093]

## Notes for your reviewers

[DT-2093]: https://outreach-io.atlassian.net/browse/DT-2093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,5 +1,6 @@
 name: github.com/getoutreach/stencil-golang
-## <<Stencil::Block(keys)>>
+###Block(keys)
+type: templates,extension
 modules:
   - name: github.com/getoutreach/stencil-discovery
 arguments:
@@ -193,4 +194,4 @@ arguments:
       type:
         - string
         - number
-## <</Stencil::Block>>
+###EndBlock(keys)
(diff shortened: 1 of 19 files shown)


manifest.yaml
```diff
@@ -1,5 +1,6 @@
 name: github.com/getoutreach/stencil-golang
-## <<Stencil::Block(keys)>>
+###Block(keys)
```
artemchernyak-outreach, line 2: Looks like old style comments ended up here somehow.

artemchernyak-outreach (approved):

manifest.yaml
jaredallard: yeah, i need to update it in `stencil-template-base`, going to just keep it for now

PR #99, opened 2022-09-20
fix: remove honeycomb tracing

artemchernyak-outreach:

We are ready to fully remove honeycomb as a specified tracing provider. This will allow us to finally bump up the version of `services`, `httpx`, and `gobox`.

## Jira ID

[XX-XX]

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -72,7 +72,6 @@ arguments:
     schema:
       type: string
       enum:
-        - honeycomb
         - opentelemetry
         - lightstep
     descriptions: Tracing backend to send traces to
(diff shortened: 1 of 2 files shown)


jaredallard (approved):

jaredallard:
@artemchernyak-outreach Breaking change entry for this?

artemchernyak-outreach:
I don't think this will actually break anything. The service should just switch over to otel transparently. There is nothing to do.

artemchernyak-outreach:
I will still add a breaking change not since it's a big switch. Just no impact or action required.

PR #109, opened 2022-09-27
feat: run multiple gRPC servers

dundee:

RMS is migrating to new version of ProtoAPI which introduces multiple gRPC servers. One for each resource (entity).

Current `rpc.go` supports only one server, therefore we introduced some manual changes in RMS to be able to run them:
https://github.com/getoutreach/rms/pull/393/files#diff-3d424a0b91fe4d66eec3c06ab49b4dacc2cd1eac42f979637f5c5153bfe74ca7R78
but we would like to solve it more consistently.

This PR is trying to handle the change in backwards-compatible way.

Also discussed here: https://outreach-hq.slack.com/archives/CM3TW594L/p1663837485326999?thread_ts=1663705133.118849&cid=CM3TW594L

## Jira ID

[XX-XX]


templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -109,7 +119,7 @@ func (s *GRPCService) Close(ctx context.Context) error {
 }
 
 // StartServer starts a RPC server with the provided implementation.
-func StartServer(ctx context.Context, service api.Service, opts... grpcx.ServerOption) (*grpc.Server, error) {
+func StartServers(ctx context.Context, servers *Servers, opts... grpcx.ServerOption) (*grpc.Server, error) {
```
jaredallard, line 122: This will be breaking because you're removing `StartServer`. Can we add back `StartServer` that just calls `StartServers` in a backwards compatible way and add a deprecation warning to the function?

dundee: Good point. I will add it.

templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -108,8 +119,13 @@ func (s *GRPCService) Close(ctx context.Context) error {
 	return nil
 }
 
-// StartServer starts a RPC server with the provided implementation.
-func StartServer(ctx context.Context, service api.Service, opts... grpcx.ServerOption) (*grpc.Server, error) {
+// StartServer is deprecated. Call StartServers instead.
```
jaredallard, line 122:
```suggestion
// StartServer is a wrapper for StartServers for compatibility reasons
// Deprecated: Call StartServers instead.
```

jaredallard (approved):

PR #112, opened 2022-09-30
fix: Exclude panic alert from devenv

huizhou-outreach:
We got a P1 alert caused by a panic from devenv. https://app.datadoghq.com/logs?cols=%40level%2Cservice%2C%40or.voice_trace_id%2C%40usr.email%2C%40k8s_namespace&event&from_ts=1664574860854&index=&live=false&messageDisplay=inline&query=panic+-%22Workflow+panic%22+status%3Aerror+service%3Atwiliowebhook+-env%3Adevelopment&stream_sort=time%2Casc&to_ts=1664575402855&viz=stream

We want to exclude those alerts.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/monitoring/datadog.tf.tpl
+++ b/templates/monitoring/datadog.tf.tpl
@@ -129,7 +129,7 @@ resource "datadog_monitor" "available_pods_low" {
 resource "datadog_monitor" "panics" {
   type    = "log alert"
   name    = "{{ .Config.Name | title }} Service panics"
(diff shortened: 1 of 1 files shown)


jaredallard:
Hmmmm, this is should already be being filtered. This is probably an issue with the Loft cluster instead, we should be just filtering on `env:development`, not namespaces as well. Can you start a thread in our support channel about this so we can get to the bottom of it instead? Thank you!! ❇️

PR #116, opened 2022-10-05
fix: Remove deprecated linters.

jkinkead:

These linters are deprecated in favor of the "unused" linter, which is already enabled.


kaldorn (approved):
Removes deadcode, structcheck, and varcheck.

jaredallard (approved):

PR #119, opened 2022-10-12
fix(linters): Disable the timeCmpSimplify linter.

jkinkead:

`timeCmpSimplify` recommends that you replace `!time.Before()` with `time.After()`, but these are not actually compatible (strictly-less being replaced with strictly-greater). Same with `!time.After()` and `time.Before()`.


jaredallard (approved):
I agree with this, but would like to see us bring this to golangci-lint as well

jkinkead:
Nice, original issue is now fixed! This override will eventually not be needed.

I'll add a TODO to remove this later, then merge.

PR #121, opened 2022-10-12
feat: on an error exit with a code

kaldorn:
If the service shuts down because of an error it should exit with a code.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -126,5 +126,7 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
 
 	if err := async.RunGroup(acts).Run(ctx); err != nil {
 		log.Warn(ctx, "shutting down service", events.NewErrorInfo(err))
+    // If there is an error it should exit with a code to indicate it
+		defer os.Exit(1)
 	}
 }
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -45,7 +45,7 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
   exitCode := 1
   defer os.Exit(exitCode)
 
-	ctx, cancel := context.WithCancel(context.Background())
+  ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
 	env.ApplyOverrides()
@@ -127,8 +127,11 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
 {{ file.Block "startup" }}
 	// <</Stencil::Block>>
 
-	if err := async.RunGroup(acts).Run(ctx); err != nil {
-		log.Warn(ctx, "shutting down service", events.NewErrorInfo(err))
+	if err := async.RunGroup(acts).Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
+		log.Error(ctx, "shutting down service", events.NewErrorInfo(err))
+		return
 	}
-  exitCode = 0
+
+	exitCode = 0
+	log.Info(ctx, "graceful shutdown process successful")
 }
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -43,8 +43,10 @@ import (
 // main is the entrypoint for the {{ .Config.Name }} service.
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
   exitCode := 1
-  defer os.Exit(exitCode)
-
+	defer func() {
+		os.Exit(exitCode)
+	}()
+	
   ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -42,11 +42,12 @@ import (
 
 // main is the entrypoint for the {{ .Config.Name }} service.
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
-	ctx, cancel := context.WithCancel(context.Background())
   exitCode := 1
   defer func() {
     os.Exit(exitCode)
   }
+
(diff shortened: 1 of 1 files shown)


templates/cmd/main.go.tpl
```diff
@@ -126,5 +126,7 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
 
 	if err := async.RunGroup(acts).Run(ctx); err != nil {
 		log.Warn(ctx, "shutting down service", events.NewErrorInfo(err))
+    // If there is an error it should exit with a code to indicate it
+		defer os.Exit(1)
```
george-e-shaw-iv, line 130: We shouldn't do this. `defer` is LIFO so this will render all other defers useless. Let me show you the right pattern

george-e-shaw-iv:
```go
func main(){
  exitCode := 1
  defer func(){
    os.Exit(exitCode)
  }()
  
  // .. do things, defer things, whatever
  
  // if we make it to the end of func main without erroring and returning we can assume success
  exitCode = 0
}
```

george-e-shaw-iv: That might not fit into what we have right now, but I hope the idea is there and it can be adapted. The important part is that this `defer` that contains the `os.Exit` call needs to be the **first** `defer` statement in `func main`

templates/cmd/main.go.tpl
```diff
@@ -43,6 +43,10 @@ import (
 // main is the entrypoint for the {{ .Config.Name }} service.
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
 	ctx, cancel := context.WithCancel(context.Background())
+  exitCode := 1
+  defer func() {
+    os.Exit(exitCode)
+  }
 	defer cancel()
```
george-e-shaw-iv, line 50:
```suggestion
  exitCode := 1
  defer func() {
    os.Exit(exitCode)
  }

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
```
ordering and keeping related items close to eachother

templates/cmd/main.go.tpl
```diff
@@ -42,6 +42,11 @@ import (
 
 // main is the entrypoint for the {{ .Config.Name }} service.
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
+  exitCode := 1
+  defer func() {
+    os.Exit(exitCode)
+  }
```
jaredallard, line 48:
```suggestion
  defer os.Exit(exitCode)
```

I think this is ok, right @george-e-shaw-iv?

templates/cmd/main.go.tpl
```diff
@@ -127,4 +132,5 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
 	if err := async.RunGroup(acts).Run(ctx); err != nil {
 		log.Warn(ctx, "shutting down service", events.NewErrorInfo(err))
 	}
+  exitCode = 0
```
jaredallard, line 135: This will always be hit? Maybe `return` in L133?

templates/cmd/main.go.tpl
george-e-shaw-iv: Yeah

templates/cmd/main.go.tpl
george-e-shaw-iv: it won't "always be hit" because initialization errors should shallow return but i do agree we should return and most likely error on 133 as opposed to just warn

we should, if `Run` doesn't already, filter out the context cancellation error though

jaredallard: * We should check if the error returned by RunGroup(acts).Run(ctx) is context.Canceled, if so, don't trigger the `log.Warn`
* Change the `log.Warn` to an error, this is an error and then immediately return (so exitCode is still 1)
* Add a log below the current conditional saying the service is shutting down (normally, since that'd be a context.Canceled or natural shutdown

george-e-shaw-iv: yeah that encapsulates the conversation + i would just say as a nit for the log below the conditional maybe say sometihng like
> graceful shutdown process successful

templates/cmd/main.go.tpl
jkinkead: This is heavily obfuscated imperative programming. The exit code is far removed from what triggers it, and you're making readers go back to the start of the function to understand why an assign is non-pointless. Much more readable is something functional:

```golang
func main() {
  exitCode := runMain()
  os.Exit(exitCode)
}

func runMain() int {
  if err != nil {
    // Stuff is bad.
    return 1
  }
  // Stuff is good.
  return 0
}
```

george-e-shaw-iv: While I appreciate your view here and I largely agree with it this is one case that I choose to go against my own values on. This is one of those things you implement exactly once and then forget about forever. It never needs to change and it always will work. Just leave a comment to make sure nobody ever puts code above the defer in func main.

george-e-shaw-iv: If we did go with the pattern you just suggested I'd much rather `runMain` return an error instead of an exit code though

george-e-shaw-iv: I've used both patterns. I like the "bad" one more (the one I suggested) personally. I won't die on this hill though.

jkinkead: Returning an error is fine - although in the general case it should include an exit code value, for when you want multiple exit codes in e.g. a CLI.

The other argument against this pattern is it makes `main` untestable, since you've called `os.Exit`. You obviously need this somewhere, and the scope where you lose coverage should be as small as possible.

kaldorn: Wouldn't it be a breaking change to have `main` return a code for this?

jkinkead: Yes, it's not in-scope for this PR (or for a quick-fix).

jaredallard (approved):

PR #122, opened 2022-10-13
fix: avoid deleting files when configs change

artemchernyak-outreach:

Deleting these files can cause issues when repos have decided to add custom version of them. This PR skips generating them, but avoids deleting.

## Jira ID

[XX-XX]


kaldorn (approved):

jkinkead:
Thanks, this broke `lintroller` (but we worked around it).

PR #129, opened 2022-10-27
fix: remove dependency from stencil-actions

tonymbugua-outreach:
Remove dependency from `stencil-actions` 

## Jira ID

DT-2925


jaredallard (approved):
LGTM, but a test is failing

PR #142, opened 2022-12-07
feat(monitoring): change http_latency_high threshold to configurable (DT-3225)

christiezhao-outreach:
## Jira ID

[DT-3225]

## Notes for your reviewers

[DT-3225]: https://outreach-io.atlassian.net/browse/DT-3225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -167,6 +167,18 @@ arguments:
     schema:
       type: number
     description: Overrides the high traffic threshold in the "http_success_rate_low" terraform module in monitoring/datadog.tf.
+  terraform.datadog.http.latency.count.highCount:
+    schema:
+      type: number
+    description: Overrides the high count threshold in the "http_latency_high" terraform module in monitoring/datadog.tf.
+  terraform.datadog.http.latency.thresholds.lowTraffic:
+    schema:
+      type: number
+    description: Overrides the low traffic threshold in the "http_latency_high" terraform module in monitoring/datadog.tf.
+  terraform.datadog.http.latency.thresholds.highTraffic:
+    schema:
+      type: number
+    description: Overrides the high traffic threshold in the "http_latency_high" terraform module in monitoring/datadog.tf.
   terraform.datadog.pods.thresholds.availableLowCount:
     schema:
       type: number
--- a/templates/monitoring/datadog.tf.tpl
+++ b/templates/monitoring/datadog.tf.tpl
@@ -166,6 +166,21 @@ variable http_success_rate_evaluation_window {
   default = {{ stencil.Arg "terraform.datadog.http.evaluationWindow" | default "last_15m" | quote }}
 }
 
+variable http_latency_high_count_threshold {
+  type = number
+  default {{ stencil.Arg "terraform.datadog.http.latency.count.highCount" | default 1000 }}
+}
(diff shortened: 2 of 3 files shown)


templates/monitoring/datadog.tf.tpl
```diff
@@ -166,6 +166,21 @@ variable http_success_rate_evaluation_window {
...
+variable http_latency_high_low_traffic_threshold {
+  type = number
+  default {{ stencil.Arg "terraform.datadog.http.latency.thresholds.lowTraffic" | default 2 }}
+}
+
+variable http_latency_high_high_traffic_threshold {
+  type = number
+  default {{ stencil.Arg "terraform.datadog.http.latency.thresholds.highTraffic" | default 2 }}
+}
+
```
jaredallard, line 183:
```suggestion
variable http_latency_high_count_threshold {
  type = number
  default = {{ stencil.Arg "terraform.datadog.http.latency.count.highCount" | default 1000 }}
}

variable http_latency_high_low_traffic_threshold {
  type = number
  default = {{ stencil.Arg "terraform.datadog.http.latency.thresholds.lowTraffic" | default 2 }}
}

variable http_latency_high_high_traffic_threshold {
  type = number
  default = {{ stencil.Arg "terraform.datadog.http.latency.thresholds.highTraffic" | default 2 }}
}

```

manifest.yaml
```diff
@@ -167,6 +167,18 @@ arguments:
     schema:
       type: number
     description: Overrides the high traffic threshold in the "http_success_rate_low" terraform module in monitoring/datadog.tf.
+  terraform.datadog.http.latency.count.highCount:
```
jaredallard, line 170: Can you add a section to the docs for these? https://github.com/getoutreach/stencil-golang/blob/main/docs/arguments.md

malept:
I think this is either a `feat` or a `fix` instead of a refactor, I'll let @jaredallard decide which is more applicable here.

jaredallard:
> I think this is either a `feat` or a `fix` instead of a refactor, I'll let @jaredallard decide which is more applicable here.

I'd say `feat`, personally!

manifest.yaml
christiezhao-outreach: Does it exist for the other variables, such as `http_success_rate_high_traffic_percentile`?

jaredallard: Ah, sorry, I should've noted. Just at the end is fine, it was missed in past PRs but just an entry _about_ the argument is good (like the other arguments)

malept:
Nit: I'm not a fan of putting the Jira issue in the semantic commit scope section. I would instead put the scope as the part of the repo or part of the code you're changing, and put the Jira ticket at the end, which would make the PR summary look like this:

```
feat(monitoring): change http_latency_high threshold to configurable (DT-3225)
```

manifest.yaml
christiezhao-outreach: Added

jaredallard (approved):
Thank you!

PR #144, opened 2022-12-21
fix: refactor how the http alerts variables are declared in service.yaml

christiezhao-outreach:
## Jira ID

[XX-XX]

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -159,14 +159,6 @@ arguments:
     schema:
       type: number
     description: Overrides the low count threshold in the "http_success_rate_low" & "http_latency_high" terraform module in monitoring/datadog.tf.
-  terraform.datadog.http.qos.thresholds.lowTraffic:
-    schema:
-      type: number
-    description: Overrides the low traffic threshold in the "http_success_rate_low" terraform module in monitoring/datadog.tf.
-  terraform.datadog.http.qos.thresholds.highTraffic:
-    schema:
-      type: number
-    description: Overrides the high traffic threshold in the "http_success_rate_low" terraform module in monitoring/datadog.tf.
   terraform.datadog.http.latency.thresholds.lowTraffic:
     schema:
       type: number
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -138,39 +138,35 @@ arguments:
   terraform.datadog.grpc.qos.thresholds.lowTraffic:
     schema:
       type: number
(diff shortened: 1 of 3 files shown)


manifest.yaml
```diff
@@ -138,39 +138,27 @@ arguments:
...
     schema:
       type: number
-    description: Overrides the high traffic threshold in the "grpc_latency_high" terraform module in monitoring/datadog.tf.
+    description: Overrides the high traffic threshold in the "grpc_success_rate_low" terraform module in monitoring/datadog.tf.
   terraform.datadog.http.evaluationWindow:
     schema:
       type: string
     description: The window of time to evaluate metrics related to HTTP events for each monitor and alert.
-  terraform.datadog.http.percentiles.lowTraffic:
+  terraform.datadog.http.qos.percentiles.lowTraffic:
```
jkinkead, line 150: These renames don't look backwards-compatible. I know we've added backcompat support before for things; maybe @jaredallard can chime in when he's back.

jkinkead (commented):
Thanks for the contribution! Overall this looks pretty good, although I'm a little worried about the breaking changes here.

From what I can tell, the main change is adding a configurable time window?

manifest.yaml
christiezhao-outreach: The main change is change the newly added http.count.highcount to lowcount and add the time window. Generally with the http latency variables added, in `service.yaml` it wasn't as clear as gRPC setup which has the success rate and latency high differentiated very clearly. But we can close this ticket if this is too big of a change for everybody!

artemchernyak-outreach: I like the addition of the time window to the metrics. But I also agree with Jesse on the rename being unnecessary.

Here are a couple options to make this mergeable:

- Remove the renaming changes, but keep the window changes.
- Keep the old names and add the new names along side. Then you can do `{{ oldName | newName }}` in the template.

We really don't want to change these stencil values unexpectedly especially now that we don't have an overarching running (like bootstrap) that can automatically change these for teams.

jkinkead: +1 to what Artem said. The new configurable value is great, and the renames are making the HTTP variables match the gRPC variables, if I read that correctly. Also great! But you should do one of the actions Artem suggested.

jaredallard:
Closing this, if you still feel like you wish to get this out @christiezhao-outreach, please feel free to take it on :D

PR #146, opened 2022-12-23
feat: dependency injection for http and grpc

george-e-shaw-iv:

* needed dependency injection for a separate private module.
* this has a breaking change of removing the deprecated StartServer
  function and then unexporting StartServers AND making it a pointer
  receiver of GRPCService. I think this is fine, I doubt anyone is
  directly calling StartServers?

## Jira ID

[STAR-915]

## Notes for your reviewers

[STAR-915]: https://outreach-io.atlassian.net/browse/STAR-915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/internal/appName/http/httpservice.go.tpl
```diff
@@ -18,6 +18,14 @@ import (
 	"github.com/getoutreach/gobox/pkg/log"
 	"github.com/getoutreach/httpx/pkg/handlers"
 
+  {{- $additionalImports := stencil.GetModuleHook "internal/http/additionalImports" }}
```
jkinkead, line 21: Will this get reformatted by `go fmt`, breaking the initial stencil run?

templates/internal/appName/http/httpservice.go.tpl
```diff
@@ -48,16 +56,37 @@ func (s *HTTPService) Run(ctx context.Context) error {
 }
 
 {{- if has "http" (stencil.Arg "serviceActivities") }}
+// PublicHTTPDependencies is used to inject dependencies into the PublicHTTPService
+// service activity.
+type PublicHTTPDependencies struct{
+    // <<Stencil::Block(publicHTTPDependencies)>>
```
jkinkead, line 62: Add a comment here explaining what a person would put into this block?

templates/internal/appName/http/httpservice.go.tpl
```diff
@@ -48,16 +56,37 @@ func (s *HTTPService) Run(ctx context.Context) error {
...
 type PublicHTTPService struct {
-	handlers.PublicService
+	  handlers.PublicService
 
-	cfg *Config
+	  cfg *Config
+    deps *PublicHTTPDependencies
 }
 
 // NewPublicHTTPService creates a new PublicHTTPService service activity.
```
jkinkead, line 84: Fix the other godocs to be period-terminated like this one, for consistency with the rest of the file, and to be inline with godoc standards?

templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -40,16 +40,37 @@ import (
 	// <</Stencil::Block>>
 )
 
+// GRPCDependencies is used to inject dependencies into the GRPCService service
+// activity.
+type GRPCDependencies struct{
+    // <<Stencil::Block(GRPCDependencies)>>
```
jkinkead, line 46: Add a comment here explaining what a person would put into this block?

templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -40,16 +40,37 @@ import (
...
 }
 
 // NewGRPCService creates a new GRPCService instance.
-func NewGRPCService(cfg *Config) *GRPCService {
-	return &GRPCService{cfg}
+func NewGRPCService(cfg *Config, deps *GRPCDependencies) *GRPCService {
+	return &GRPCService{
+        cfg: cfg,
+        deps: deps,
+    }
```
jkinkead, line 73: Funky indent here; doesn't align with `70`.

templates/internal/appName/http/httpservice.go.tpl
```diff
@@ -48,16 +56,37 @@ func (s *HTTPService) Run(ctx context.Context) error {
...
-func NewPublicHTTPService(cfg *Config) *PublicHTTPService {
-	return &PublicHTTPService{cfg: cfg}
+func NewPublicHTTPService(cfg *Config, deps *PublicHTTPDependencies) *PublicHTTPService {
```
jkinkead, line 85: Can you mark the old signature deprecated, and create a new one instead? I'd like to start giving customers some advance notice for breaking changes.

Ditto for the `StartServers` call, which [is being used](https://cs.github.com/getoutreach/kaiagateway/blob/ad1f20dbf839c9b6f346632617e2f43da7926b2b/internal/kaiagateway/bento_gateway_test.go?q=startservers+#L39).

george-e-shaw-iv: Not doing this as per the conversation we had about limited blast radius here.

templates/internal/appName/http/httpservice.go.tpl
george-e-shaw-iv: We already do this in various other places, like in `rpc.go`, so I think we're good here?

templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -40,16 +40,38 @@ import (
...
 // being implemented is a gRPC server.
 type GRPCService struct {
 	cfg *Config
+	deps *GRPCDependencies
 }
 
 // NewGRPCService creates a new GRPCService instance.
-func NewGRPCService(cfg *Config) *GRPCService {
-	return &GRPCService{cfg}
+func NewGRPCService(cfg *Config, deps *GRPCDependencies) *GRPCService {
```
samnguyen-outreach, line 70: Can we (i.e. is it possible to) leave the existing signature unmodified and introduce a new one to avoid a breaking change? It would a nicer user experience to have a deprecation period instead of, say, running stencil from some other reason and then being forced to refactor stuff (unless the refactor ends up being trivial)

```
func NewGRPCServiceWithDependencies(cfg *Config, deps *GRPCDependencies) *GRPCService {
```


(same comment applies to HTTPService)

george-e-shaw-iv: This breaking change will only affect 2 services,`mint` and `polaroid`, which is why we opted to *not* do what you're proposing (Jesse already proposed it and we talked about it during a design review meeting this morning).

artemchernyak-outreach (approved):

PR #152, opened 2023-01-12
feat: allow overriding Datadog SLOs

kaldorn:
Setting this as a static file will allow the monitors to be updated/changed as each service requires.

## Jira ID

[DT-3230]

## Notes for your reviewers

[DT-3230]: https://outreach-io.atlassian.net/browse/DT-3230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/monitoring/datadog.tf.tpl
+++ b/templates/monitoring/datadog.tf.tpl
@@ -1,5 +1,5 @@
 {{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}
-{{- $_ := file.Static }}
+{{- $_ := file.Static }} 
 variable P1_notify {
   type = list(string)
   default = []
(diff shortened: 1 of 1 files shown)


templates/monitoring/datadog.tf.tpl
```diff
@@ -1,4 +1,5 @@
 {{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}
+{{- $_ := file.Static }}
```
jkinkead, line 2: I don't think we want the entire Datadog monitoring file to be user-managed. This would dramatically reduce the ability of us to roll out new features.

I was thinking more along the lines of:
* SLO block for people to add custom SLOs.
* boolean to opt-out of default SLOs.

PR #165, opened 2023-01-24
fix: missing data should not set off monitors

kaldorn:
Setting the default value to zero will present missing data as `0`, which is how we want to interpret it so that our monitors and SLOs don't fire when data is missing.

## Jira ID

[DT-3381]

## Notes for your reviewers

[DT-3381]: https://outreach-io.atlassian.net/browse/DT-3381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/monitoring/datadog.tf.tpl
+++ b/templates/monitoring/datadog.tf.tpl
@@ -230,9 +230,8 @@ module "http_success_rate_low" {
   Notify: ${join(" ", var.P2_notify)}
   EOF
   require_full_window = false
-  low_count_query = "sum(${var.http_success_rate_evaluation_window}):count:${local.http_request_seconds}{!status:5xx,!env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count() < ${var.http_success_rate_low_count_threshold}"
-  low_traffic_query = "sum(${var.http_success_rate_evaluation_window}):100 * ( count:${local.http_request_seconds}{!status:5xx,!env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count() / count:${local.http_request_seconds}{*, !env:development} by {kube_namespace}.as_count() ) < ${var.http_success_rate_low_traffic_percentile}"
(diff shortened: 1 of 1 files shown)


artemchernyak-outreach (approved):

kaldorn:
> This doesn't fix the SLOs per the ticket.

There isn't a way to modify the query that the SLOs have. The query is fixed on the monitors that the SLO is generating uptime on.

jkinkead (commented):
Specifically, this misses the success rate SLOs.

jkinkead:
(maybe those are always fine given they're a count, but I think I've seen them with empty data for some services)

jkinkead:
[Delibird example](https://app.datadoghq.com/slo/manage?query=%22Success%20Response%22&slo_id=559e51bb9a2255468161978d10c67414&tab=status_and_history&timeframe=7d).

templates/monitoring/datadog.tf.tpl
```diff
@@ -277,8 +277,8 @@ resource "datadog_service_level_objective" "http_success" {
   description = "Comparing 5xx responses to all requests as a ratio, broken out by bento."
   tags = local.ddTags
   query {
-    numerator   = "count:${local.http_request_seconds}{!status:5xx, !env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count()"
-    denominator = "count:${local.http_request_seconds}{*, !env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count()"
+    numerator   = "count:${local.http_request_seconds}{!status:5xx, !env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count().fill(zero)"
+    denominator = "count:${local.http_request_seconds}{*, !env:development,app:{{ stencil.ApplyTemplate "goPackageSafeName" }}} by {kube_namespace}.as_count().fill(zero)"
```
jkinkead, line 281: Won't this still cause issues like this one?
image

kaldorn: 100% it does.

jkinkead (approved):
Approving to unblock; we've talked through all the issues. :)

PR #168, opened 2023-01-30
feat: add sync module hook

artemchernyak-outreach:

Adds a module hook that allows expanding the devbase sync.

## Jira ID

[DT-3309]

## Notes for your reviewers

[DT-3309]: https://outreach-io.atlassian.net/browse/DT-3309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


masterkidan (approved):
LGTM, adds module hook for sync

kaldorn (approved):

templates/devspace.yaml.tpl
```diff
@@ -157,6 +157,18 @@ dev:
         {{- if (has "node" (stencil.Arg "grpcClients")) }}
         - api/clients/node/node_modules/
         {{- end }}
+    {{- range (stencil.GetModuleHook "devspace.sync") }}
+    - name: {{ .name }}
```
jaredallard, line 161: Doesn't this significantly muddy the move to devspace v2? I'm... a little worried about that. Though I guess we probably already have module hooks here.

artemchernyak-outreach: Yeah we have a few module hooks already and `stencil-smartstore` is using those as well. The new filesync syntax seems to be mostly compliant with this, so we may not even have to change the hooks themselves. That was why I used all these variables here.

New syntax docs: https://www.devspace.sh/docs/configuration/dev/connections/file-sync

PR #170, opened 2023-01-30
feat: add module hook to add make commands

tmerkel:
Adds a module hook to allow other modules to add Make commands to the root Makefile.

Allows `stencil-graphql` to add a command for running the rover-cli for dev work.

## Jira ID

[XX-XX]

### DIFF
--- a/docs/module-hooks.md
+++ b/docs/module-hooks.md
@@ -198,10 +198,14 @@ Extra ports that should be forwarded while `devspace dev` is running
 Extra commands to add to the root Makefile
 
 ```tpl
+{{- define "run.rover" }}
 ## run-rover:           merges shared and specific schemas and runs rover-cli 
 .PHONY: run-rover
 run-rover:
-    cat internal/graphql/schema/shared.graphql > internal/graphql/generated/schema.graphql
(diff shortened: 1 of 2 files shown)


docs/module-hooks.md
```diff
@@ -175,3 +175,20 @@ Extra [devspace profiles](https://www.devspace.sh/docs/5.x/configuration/profile
...
+
+### `Makefile.commands`
+
+**Type**: `string`
+
+**File**: `Makefile`
+
+Extra commands to add to the root Makefile
+
+```tpl
```
artemchernyak-outreach, line 187: This example should include an actual sample of how you would apply this. This should include the call to `stencil.AddToModuleHook`. 

Ideally, this example should come from your test case of running this inside a stencil module. You can always use `replacement` inside `service.yaml` to test this locally.

tmerkel: ...I failed to capture the surrounding context :(. Thanks.  Let me try that again.

tmerkel: Better?

artemchernyak-outreach: Perfect! Thanks for expanding this.

artemchernyak-outreach (approved):

PR #179, opened 2023-02-07
feat: added ChannelOptions param to client-helpers ClientOptions

jdidis:

This change allows developers to pass in [ChannelOptions](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/channel-options.ts#L23) during node client initialization.

## Jira ID

[XX-XX]


tonymbugua-outreach (approved):
LGTM - adds channel options param

PR #180, opened 2023-02-07
fix:  devspace handle multiple containers

kaldorn:
Have devspace default to the container named after the service/app to handle multiple container pods. This should handle most use cases.

## Jira ID

[DT-3136]

## Notes for your reviewers

[DT-3136]: https://outreach-io.atlassian.net/browse/DT-3136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/devspace.yaml.tpl
+++ b/templates/devspace.yaml.tpl
@@ -148,6 +148,7 @@ dev:
   # `dev.sync` configures a file sync between our Pods in k8s and your local project files
   sync:
     - name: app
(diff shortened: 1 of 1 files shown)


jaredallard (approved):
Seems legit, aligns with my understanding of our jsonnet as well.

PR #182, opened 2023-02-07
feat: default go mod to golang 1.19

kaldorn:
Default to 1.19

## Jira ID

[XX-XX]


jaredallard (approved):
Updating go version to 1.19 because it doesn't make much sense to gate generics anymore. sgtm

PR #185, opened 2023-02-15
fix: enable concurrent reconciles

evadnoob:
Enables concurrent reconciles, also requires changes in services

https://github.com/getoutreach/services/pull/382


templates/internal/appName/k8s/activity.go.tpl
```diff
@@ -66,8 +67,9 @@ func (s *KubernetesService) Run(ctx context.Context) error { //nolint: funlen,ll
 	ctrl.SetLogger(logadapters.NewLogrLogger(ctx))
 
 	s.registerSchemes()
-
-	options := ctrl.Options{
+  
+    // manager options
+    mgrOptions := ctrl.Options{
```
AndrewWinterman, line 72: can we use the whole word? `manager` was not the first thing i thought of

AndrewWinterman (approved):
appears to be updating stencil based on updated apis in Services

templates/internal/appName/k8s/activity.go.tpl
evadnoob: Ok, changed

AndrewWinterman (approved):
still looks fine

jaredallard (approved):
Seems solid. Will review the larger `services` part soon

PR #193, opened 2023-02-22
feat: remove trash monitor

george-e-shaw-iv:

* removes the trash (avaliable pods low) monitor from templates
* the reason this is trash is due to:
  * it is sensitive to deployments
  * it is sensitive to normal rescheduling onto new nodes
  * it doesn't account for the fact that different enviroments
    can and often do have different base pod counts
  * etc

## Jira ID

[XX-XX]


pbr0ck3r (approved):

jaredallard (approved):
Fundamentally I agree with this, pod cycling is _not_ a deterministic indicator that an application is going to have issues. As such, I think it makes sense to remove this. What we provide should be high-quality for all developers.

jkinkead (commented):
A useful alert would be: Notify me if my pods drop below my disruption budget.

Can you open an issue tracking adding an "exceeded disruption budget" monitor to this template?

jaredallard:
> A useful alert would be: Notify me if my pods drop below my disruption budget.
> 
> Can you open an issue tracking adding an "exceeded disruption budget" monitor to this template?

I'm going to be handling default alerts as part of my greater work here. I think that covers it, right?

PR #199, opened 2023-02-28
fix: correct vscode debugger path mapping

huizhou-outreach:

After we added the build flag `trimpath`, instead of absolute file system paths, the recorded file names will begin either a module path@version (when using modules), or a plain import path (when using the standard library, or GOPATH).

We need to update `substitutePath` in vscode debug settings so that the debugger can find the proper code path.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/.vscode/launch.json.tpl
+++ b/templates/.vscode/launch.json.tpl
@@ -31,7 +31,11 @@
       "substitutePath": [
         {
           "from": "${workspaceRoot}",
-          "to": "/home/dev/app"
+          "to": "github.com/getoutreach/{{ .Config.Name }}"
+        },
+        {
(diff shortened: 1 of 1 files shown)


jaredallard:
@huizhou-outreach thank you!! Mind adding some context to the description so we remember what this was for? Thank you!

huizhou-outreach:
@jaredallard updated.

pranavkedia (approved):

huazhoul-outreach (approved):

jaredallard (approved):
The second rule seems like a duplicate of the one below it, but we can see what happens 🤷

huizhou-outreach:
TBH, I don't know what's `/tmp/cache/go/mod/` used for. Just keep it.

jkinkead (commented):
Awesome, thanks!

PR #200, opened 2023-02-28
fix(DT-3543): Cleaning up out-of-the-box imports

deregtd:
Also had to add a new imports module hook to handle stuff from:
https://github.com/getoutreach/stencil-outreach/pull/113/files

### DIFF
--- a/templates/internal/appName/config.go.tpl
+++ b/templates/internal/appName/config.go.tpl
@@ -10,13 +10,12 @@ package {{ stencil.ApplyTemplate "goPackageSafeName" }} //nolint:revive // Why: 
 
 import (
 	"context"
-	"os"
-	"strings"
 
 	"github.com/getoutreach/gobox/pkg/cfg"
 	"github.com/getoutreach/gobox/pkg/log"
-	"github.com/getoutreach/gobox/pkg/events"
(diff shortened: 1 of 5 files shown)


templates/internal/appName/config.go.tpl
```diff
@@ -10,13 +10,12 @@ package {{ stencil.ApplyTemplate "goPackageSafeName" }} //nolint:revive // Why:
...
 import (
 	"context"
-	"os"
-	"strings"
 
 	"github.com/getoutreach/gobox/pkg/cfg"
 	"github.com/getoutreach/gobox/pkg/log"
-	"github.com/getoutreach/gobox/pkg/events"
+	{{- if has "kafka" (stencil.Arg "serviceActivities") }}
 	"github.com/getoutreach/services/pkg/find"
```
george-e-shaw-iv, line 17: This isn't an OSS repo btw (not ur fault but I see it so I'll call it out)

deregtd: yeah ugh we'll have to add a hook here.  i'll open a ticket.  that'll be more invasive surgery.

jaredallard (approved):

PR #213, opened 2023-03-02
fix(tests): stop generating all_tests.go

jaredallard:

This PR stops generating `all_test.go` (but does not remove it). `all_test.go` was a way to trigger test shuffling, which is no longer needed and is deprecated in the latest `gobox`. We stop generating this file to prevent new services from getting deprecated, unchangable, code. We leave it for existing services so they can handle the deprecation as needed on their side, while still letting tests work and run as normal.

## Jira ID

[XX-XX]


kaldorn (approved):

kaldorn:
Sleep well, all_test.go

PR #216, opened 2023-03-07
feat: added ModuleHook for dependencyInjection

bhogatemadhura:
We already have a stencil block for the services to add more dependency. Need a way to add more dependency for stencil modules

## Jira ID

https://outreach-io.atlassian.net/browse/STAR-1200

### DIFF
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -57,15 +57,11 @@ type dependencies struct{
   {{- if has "grpc" (stencil.Arg "serviceActivities") }}
   gRPC {{ $pkgName }}.GRPCDependencies
   {{- end }}
-  {{- $maindependencies := stencil.GetModuleHook "main.dependencies" }}
-  {{- if $maindependencies }}
-  // additional dependencies injected by modules
-    {{- range $maindependencies }}
-  {{ . }}
-    {{- end }}
-  // end additionaldependencies injected by modules
+  {{- range stencil.GetModuleHook "main.dependencies" }}
+  {{- range $k, $v := . }}
+  {{ $k }} {{ $v }}
   {{- end }}
-
+  {{- end }}
   // <<Stencil::Block(customServiceActivityDependencyInjection)>>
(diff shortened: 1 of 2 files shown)


jaredallard (commented):
Please add documentation here https://github.com/getoutreach/stencil-golang/blob/main/docs/module-hooks.md, it makes it easier for people to consume the new module hook 🙇

templates/cmd/main.go.tpl
```diff
@@ -57,6 +57,14 @@ type dependencies struct{
   {{- if has "grpc" (stencil.Arg "serviceActivities") }}
   gRPC {{ $pkgName }}.GRPCDependencies
   {{- end }}
+  {{- $additionaldependencies := stencil.GetModuleHook "additionaldependencies" }}
```
jaredallard, line 60:
```suggestion
  {{- $additionaldependencies := stencil.GetModuleHook "main.dependencies" }}
```

Suggestion on an alternative name that's also better namespaced.

templates/cmd/main.go.tpl
```diff
@@ -57,6 +57,14 @@ type dependencies struct{
   {{- if has "grpc" (stencil.Arg "serviceActivities") }}
   gRPC {{ $pkgName }}.GRPCDependencies
   {{- end }}
+  {{- $maindependencies := stencil.GetModuleHook "main.dependencies" }}
+  {{- if $maindependencies }}
+  // additional dependencies injected by modules
+    {{- range $maindependencies }}
+  {{ . }}
```
jaredallard, line 64: Sorry to bike shed on this, but after having a convo tangentially related, it'd be better to prescribe a certain format here.

Maybe we could do `[]string{}` or something more strongly typed, such as:

```go
map[string]string{
  "field": "myFieldName",
  "type": "myType"
}
```

This would make the API provided by the module hook stronger and less likely to break in the future. If we did that, making the name `main.dependencyFields` might make more sense as well.

We don't have to do this immediately, but it would be good to be tracked on the DT side if we don't to prevent accidental format breakage later

jaredallard (approved):

PR #225, opened 2023-03-17
fix: support python, dash separated repo

ygcao:
- tried to unify the serviceActivities definition location, to only one place in stencil-base, but found it causes demands on new parameters, so simply added python into the existing list here
- fix the long standing pain which forces people to use ugly repo names without any symbols, grpc code doesn't make name right
- add hook for insert new images from other modules, not sure how the gitaction part is used but not fit our simple needs
- skip Go traffic handler code generations for `http` and `grpc` when `python` is turned on, we'll make skeleton codes instead to proxy Python service behind. This helps to isolate our codes and minimize the needs of your efforts for review,  instead of trying to hack by lots of `if else` statements in your codes from time to time when we try to keep evolving more.

## Jira ID

[[DSC-10572](https://outreach-io.atlassian.net/browse/DSC-10572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)]

## Notes for your reviewers

Tested locally, all worked as desired. Will follow up with the build error before merge if any.

[DSC-10572]: https://outreach-io.atlassian.net/browse/DSC-10572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -181,8 +181,8 @@ func (gs *GRPCService) StartServers(ctx context.Context, servers *Servers, opts.
 	// end gRPC RPCs injected by modules
 	{{- end }}
 
-	// Register default server
-	api.Register{{ title $pkgName }}Server(s, rpcserver{servers.DefaultServer})
+	// Register default server, title function won't work well when use underscore, so we make it dash first
+	api.Register{{ $pkgName | replace "_" "-" | title | replace "-" "" }}Server(s, rpcserver{servers.DefaultServer})
```
jaredallard, line 185: Use `goPackageSafeName` function here instead?

manifest.yaml
```diff
@@ -32,7 +32,8 @@ arguments:
           - grpc
           - temporal
           - kafka
-    description: Any valid built-in service activities can be supplied here and they will be provided. The `service` flag must be set to true. Valid values for this are `http`, `grpc`, `temporal`, and `kafka`.
+          - python
+    description: Any valid built-in service activities can be supplied here and they will be provided. The `service` flag must be set to true. Valid values for this are `http`, `grpc`, `temporal`, `python` and `kafka`.
```
jaredallard, line 36: Instead of putting `python` here, I think this should be scoped to the stencil-python module or w/e you're calling it for Outreach.

Otherwise, we're polluting this package.

jaredallard (changes requested):

templates/internal/appName/rpc/rpc.go.tpl
ygcao: It won't work, pkgName is already from it.
```{{- $pkgName := stencil.ApplyTemplate "goPackageSafeName" }}```

File: manifest.yaml (hunk 2 above)
Comment (jaredallard) [reply manifest.yaml:36]: I'm not trying to hate on python, I meant that it shouldn't be inside of `stencil-golang` because it's scope isn't python applications.

Comment (ygcao) [reply manifest.yaml:36]: could you look at the usage? any other `working` choice?

File: templates/internal/appName/rpc/rpc.go.tpl (hunk 1 above)
Comment (jaredallard) [reply templates/internal/appName/rpc/rpc.go.tpl:185]: Ah, I see

File: manifest.yaml (hunk 2 above)
Comment (ygcao) [reply manifest.yaml:36]: @deregtd any comment on this? We discussed adding `python` as yet another `serviceActivity` in [slack thread](https://outreach-hq.slack.com/archives/C01CAR6GVPD/p1678316708202279) already.  I've already abandoned the possibility of `python-grpc` since found one type is good enough when combining with either `http` or `grpc` for our decision makings.

Comment (deregtd) [reply manifest.yaml:36]: The key would be figuring out extensions points to stencil-golang to allow stencil-datascience to extend an automated GRPC interface on top of the basic golang stuff, and then insert a new container into stencil-base/devbase, etc.  stencil-golang should have no idea python exists, just as if we had stencil-python, it should have no idea stencil-golang exists, you'd be working with the base modules.  You're actually trying to do a specialization where you add MORE on top of stencil-golang, including adding a custom build of a python module, so you really want to compose both the idea of stencil-golang for the base service as well as a new second container on top of it.  So I'd suggest doing stencil-mlwrapper or stencil-wrappedpython or something like that.  stencil-python would be something like the analogy to stencil-golang, but for a python service as the only service getting deployed.  You're doing a hybrid, if that makes sense.

File: templates/internal/appName/rpc/rpc.go.tpl (hunk 1 above)
Comment (ZhaopuTeng) [inline templates/internal/appName/rpc/rpc.go.tpl:185]: @ygcao, I believe a few more place also need to have similar changes. 
One example:
https://github.com/getoutreach/stencil-golang/blob/3f5fd3cef6ff980546f2a6b859ed4588d88ec812/templates/api/rpc/client.go.tpl#L75
And also L83, L105, L114.

Comment (ygcao) [reply templates/internal/appName/rpc/rpc.go.tpl:185]: you could be right. I didn't try the Go client yet thus didn't bother by it yet. will add same logic there too. maybe it's time for making it another function to use in needed places then. something like `goTitleCaseName`

File: manifest.yaml (hunk 2 above)
Comment (deregtd) [reply manifest.yaml:36]: Shouldn't that be module hooks within stencil-go, so that you can poke into it from other modules?

Comment (ygcao) [reply manifest.yaml:36]: It was difficult since I have to change a lot, after refactoring, it may be possible now. Still module hooks has less flexibility than the entire overwrite when we have new needs later. 
FYI: this is the motivation I mentioned in description for why I finally chose to do skipping instead of hacking into a given section. Now, it seems fine for the hook approach, but I don't know about the future. If you still have strong opinion here for using hooks after reading following and don't bother to get more PRs in the future about potential more customizations, I can go back to the original thinking.
```
skip Go traffic handler code generations for http and grpc when python is turned on, we'll make skeleton codes instead to proxy Python service behind. This helps to isolate our codes and minimize the needs of your efforts for review, instead of trying to hack by lots of if else statements in your codes from time to time when we try to keep evolving more.
```

File: templates/internal/appName/rpc/rpc.go.tpl (hunk 1 above)
Comment (ygcao) [reply templates/internal/appName/rpc/rpc.go.tpl:185]: I finally found I indeed has the go client code generated, and there were fine somehow. Looks doesn't need a fix, do you have example broken generated codes?
image

Comment (ygcao) [reply templates/internal/appName/rpc/rpc.go.tpl:185]: Never mind, found since the file was `Static`, it didn't do refresh for the new runs of `stencil`. They indeed need a fix too.

Comment (ZhaopuTeng) [reply templates/internal/appName/rpc/rpc.go.tpl:185]: Thanks for the confirmation. FYI, this is a test from my side: 
Screenshot 2023-03-20 at 11 57 51 AM

File: manifest.yaml (hunk 2 above)
Comment (ygcao) [reply manifest.yaml:36]: Also noticed that even when I switch to use Module hook to get rid of the usage of python `serviceActivities`, since `stencil-base` [still need it](https://github.com/getoutreach/stencil-base/pull/100/files#diff-885b271153664f100985178e148d28687d6220474d93e4526bf5f9e17c5263d9R60), we can't simply remove `python` from this repo since it will cause `stencil` to complain on validation error since we still need to feed `python` for `stencil-base`'s usage in the generated repo.

Comment (ygcao) [reply manifest.yaml:36]: Summary of today's meeting with @deregtd 
1. Due to the fact that rpc flag is already being used in `stencil-base`,  python is just a similar injection, We'll keep the flag adding in stencil-base and stencil-golang as MLP proposed now. DT will find ways to refactor later and MLP can participate the changes to make it still work for python community.
2. We'll use Module hook for the injection of new Go code in this repo instead of doing swapping, although it means more code changes now and maybe in the future.
I've pushed my changes on 2 accordingly. Please review and approve. @jaredallard if you still have concerns, let us know, if needed, I like to have another 1:1 meeting with you on an more Prague friendly time.

File: templates/internal/appName/http/handler.go.tpl [hunk 3]
```diff
@@ -26,7 +39,10 @@ func Handler() http.Handler {
 	// Replace this with your routes
 	routes.Handle("/ping", handlers.Endpoint("ping", svc.ping))
 	routes.Handle("/pong", handlers.Endpoint("pong", svc.pong))
-
+{{- $extraRoutes := (stencil.GetModuleHook "http/extraRoutes") }}
```
jaredallard, line 42: Can you add a comment above these denoting that they were added by module hooks? We don't have a good debugability story for module hooks w/o that right now.

jaredallard (commented):
Generally LGTM right now. Can we get docs added for each of the new module hooks added into the existing module hook docs?

templates/internal/appName/http/handler.go.tpl
```diff
@@ -6,15 +6,28 @@
 // {{ stencil.ApplyTemplate "copyright" }}
 
 // Description: This file exposes the public HTTP service for {{ .Config.Name }}.
+{{- $extraComments := (stencil.GetModuleHook "http/extraComments") }}
+{{- range $extraComments }}
+{{- .}}
```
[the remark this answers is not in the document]
ygcao: just a place holder for educating users about what we'll add for the proxy behavior in our own Go logics inserted from our template.

ygcao:
> Generally LGTM right now. Can we get docs added for each of the new module hooks added into the existing module hook docs?

could you provide a link to it? I'm glad to add to it once I know where it is. Thanks!

jaredallard:
> > Generally LGTM right now. Can we get docs added for each of the new module hooks added into the existing module hook docs?
> 
> could you provide a link to it? I'm glad to add to it once I know where it is. Thanks!

https://github.com/getoutreach/stencil-golang/blob/main/docs/module-hooks.md

jaredallard (approved):

PR #229, opened 2023-03-20
feat: support unmanaged CLIs

jaredallard:

This PR adds support for unmanaged CLIs, useful for when you want to use the stencil CLI distribution framework but not opt into the CLI framework (e.g., no updater, or other aspects of it).

This is done by overloading the `commands` argument to support the original:

```yaml
commands:
- a-cool-command
```

but also a new format:

```yaml
commands:
- a-cool-command:
    option: true
```

To opt-out, one can now set `unmanaged`:

```yaml
commands:
- a-cool-command:
    unmanaged: true
```

This unmanaged the CLI file by setting it to 'static' and generating only a minimal `main.go` placeholder. Goreleaser, semantic-release, and etc are still used to package and distribute the binary as normal.

## Jira ID

[XX-XX]

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -42,7 +42,13 @@ arguments:
     schema:
       type: array
       items:
-        type: string
+        anyOf:
+          - type: string
+          - type: object
+            properties:
+              unmanaged:
+                type: boolean
+                description: If true, stencil will not generate an entrypoint go file for this CLI (`cmd/<name>/<name>.go`), but still will build and distribute it.
     description: List of CLI commands to generate for this repository
   grpcClients:
     schema:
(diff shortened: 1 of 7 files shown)


manifest.yaml
```diff
@@ -42,7 +42,13 @@ arguments:
     schema:
       type: array
       items:
-        type: string
+        anyOf:
```
chatchai-outreach, line 45: [stencil-outreach](https://github.com/getoutreach/stencil-outreach/blob/d8af105a02820d09079229b67a472ddfa71162c2/manifest.yaml#L41-L46) needs to be updated as well

chatchai-outreach (approved):
this is a great feature :D

PR #234, opened 2023-03-22
fix(release): use stencil-outreach for opslevel

jaredallard:

This PR adds a dependency on `stencil-outreach` so that we're able to release w/ `opslevel`

## Jira ID

[XX-XX]


zaquestion (approved):
also begrudging approval

PR #240, opened 2023-03-24
feat: add pagerduty p1 p2 to manifest yaml for loading into template …

kaldorn:
…files

On the tin. Adds this to the manifest.yaml and parses it into the tfvars for monitoring.

## Jira ID

[DT-1693]

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -123,6 +123,12 @@ arguments:
       items:
         type: string
     description: Dependencies your repository needs, but not always (e.g. E2E tests).
+  pagerduty.p1:
+    schema:
+      type: string
+  pagerduty.p2:
+    schema:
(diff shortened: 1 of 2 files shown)


manifest.yaml
```diff
@@ -123,6 +123,12 @@ arguments:
       items:
         type: string
     description: Dependencies your repository needs, but not always (e.g. E2E tests).
+  pagerduty.p1:
```
chatchai-outreach, line 126: I thought that we want to scrape this data from https://github.com/getoutreach/cloud-governance-terraform/blob/main/teams.json

jaredallard: +1, I think we want to load this data automatically and stop having it be set

PR #244, opened 2023-03-29
fix: Fixing controller metrics

masterkidan:

This fixes metrics collection for controller_runtime_metrics , which are published by operators.

K8s metrics are published on port 2019, as they use the metrics server from the ControllerManager. We can't bind that metrics server to the 8000 port (which we use for exporting other metrics, and is already bound).

## Jira ID

[XX-XX]


templates/deployments/appname/app.jsonnet.tpl
```diff
@@ -292,11 +292,19 @@ local all = {
 								send_distribution_buckets: true,
 							},
 						],
-						{{- if not (empty (stencil.Arg "kubernetes.groups")) }}
-						['ad.datadoghq.com/' + k8sMetricsName + '.check_names']: '["openmetrics"]',
-						['ad.datadoghq.com/' + k8sMetricsName + '.init_configs']: '[{}]',
-						['ad.datadoghq.com/' + k8sMetricsName + '.instances']: std.manifestJsonEx(self.k8s_datadog_prom_instances_, '  '),
+            {{- else }}
+						['ad.datadoghq.com/' +  app.name + '.check_names']: '["openmetrics","openmetrics"]',
```
jaredallard, line 296: Can we add a comment here about why the value is duplicated. Just worried someone will accidentally remove this later

masterkidan: done

jaredallard (approved):

PR #248, opened 2023-04-03
feat: [ORES-1910] Remove access token and auth interceptor

voslartomas:

Removing `accessToken` and auth interceptor as this forces clients to create client for each request. Instead authentication will be handled directly in Giraffe, via shared interceptor.

## Jira ID

[ORES-1910]

## Notes for your reviewers

[ORES-1910]: https://outreach-io.atlassian.net/browse/ORES-1910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/api/clients/node/src/client-helpers.ts.tpl
+++ b/templates/api/clients/node/src/client-helpers.ts.tpl
@@ -1,7 +1,7 @@
 {{- $_ := stencil.ApplyTemplate "skipGrpcClient" "node" -}}
 import * as grpc from '@grpc/grpc-js';
 import { {{ stencil.ApplyTemplate "serviceNameLanguageSafe" }}Client } from './grpc/{{ .Config.Name }}_grpc_pb';
-import { createAuthenticationInterceptor, createErrorLoggerInterceptor } from '@outreach/grpc-client';
+import { createErrorLoggerInterceptor } from '@outreach/grpc-client';
 import winston from 'winston';
 import * as find from '@outreach/find';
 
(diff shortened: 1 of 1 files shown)


jaredallard (approved):
I have no context on what these changes do, so I'm approving to satisfy the DT approval. That said, I'd ask that this change not break existing clients. If it does, then we should workshop this.

voslartomas:
> I have no context on what these changes do, so I'm approving to satisfy the DT approval. That said, I'd ask that this change not break existing clients. If it does, then we should workshop this.

It does break clients, but I am fixing it on the way in https://github.com/getoutreach/giraffe/pull/3330

I don't see any other way how to do this.

jaredallard:
> > I have no context on what these changes do, so I'm approving to satisfy the DT approval. That said, I'd ask that this change not break existing clients. If it does, then we should workshop this.
> 
> It does break clients, but I am fixing it on the way in [getoutreach/giraffe#3330](https://github.com/getoutreach/giraffe/pull/3330)
> 
> I don't see any other way how to do this.

If it breaks other clients then we should have a writeup of the actions for the user to take before this gets merged, just a simple one-pager is fine I think.

voslartomas:
> > > I have no context on what these changes do, so I'm approving to satisfy the DT approval. That said, I'd ask that this change not break existing clients. If it does, then we should workshop this.
> > 
> > 
> > It does break clients, but I am fixing it on the way in [getoutreach/giraffe#3330](https://github.com/getoutreach/giraffe/pull/3330)
> > I don't see any other way how to do this.
> 
> If it breaks other clients then we should have a writeup of the actions for the user to take before this gets merged, just a simple one-pager is fine I think.

Ok, I can come up with something tomorrow. But I believe node grpc clients are used only in Giraffe, so there is no need for other users. I am creating PR in every single service used in Giraffe and updating it.

voslartomas:
Where do you want me to send this info?

jaredallard:
@voslartomas ah, if you're going to update the services I'm less concerned. Sending something in #dev-announce would be fine with me, if you still want to do so!

jaredallard (approved):

voslartomas:
@jaredallard I think it would take ages if I would let others to update it as it is not urgent for them. So I decided to do it all at once, as this is eating memory in Giraffe and we need to do something about it really fast. I will create announce about all this

PR #284, opened 2023-05-18
fix(vscode): use proper debugger paths for dlv debug

jaredallard:

This PR fixes the local paths to no longer reply on `-trimpath`. When https://github.com/getoutreach/devbase/pull/510 landed, we stopped using `trimpath` (unobviously) which broke the debugger path substitution. While here, I added in-line code docs to make it easier to understand what each part is doing and why it exists.

I tested this locally on `keraton`.

## Jira ID

[DT-3746]

[DT-3746]: https://outreach-io.atlassian.net/browse/DT-3746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kaldorn (approved):

jaredallard:
Side note: The VSCode Go debugger docs are great https://github.com/golang/vscode-go/blob/master/docs/debugging.md

PR #292, opened 2023-05-22
fix: remove temporal pods low monitors

bhogatemadhura:
removed the temporal monitor which was using the deprecated variable `available_pods_low_count`

## Jira ID

https://outreach-io.atlassian.net/browse/STAR-1281

### DIFF
--- a/templates/.snapshots/TestBasicGoMod-go.mod.tpl-go.mod.snapshot
+++ b/templates/.snapshots/TestBasicGoMod-go.mod.tpl-go.mod.snapshot
@@ -3,5 +3,5 @@
 go 1.19
 
 require (
-	github.com/getoutreach/gobox v1.70.4
+	github.com/getoutreach/gobox v1.70.5
 ))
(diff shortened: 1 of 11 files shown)


jaredallard:
Considering we removed the monitor that initially added this, I think we should probably remove the correlating temporal one, right? 

Also, should we move [temporal.tf.tpl](https://github.com/getoutreach/stencil-golang/blob/76a6e5dbac0037964f297d421b61cc32528c0295/templates/monitoring/temporal.tf.tpl#L60) into stencil-temporal?

bhogatemadhura:
> Considering we removed the monitor that initially added this, I think we should probably remove the correlating temporal one, right?
> 
> Also, should we move [temporal.tf.tpl](https://github.com/getoutreach/stencil-golang/blob/76a6e5dbac0037964f297d421b61cc32528c0295/templates/monitoring/temporal.tf.tpl#L60) into stencil-temporal?

I'll let @ChrisPRobinson  chime in here

ChrisPRobinson:
@jaredallard ya that's fine with me, I don't really want temporal monitors in this repo anyway

ChrisPRobinson:
I'd rather have them in stencil-temporal but I either missed moving them or moving them was a bit complex so I left them where they were

george-e-shaw-iv (approved):

jaredallard (approved):

PR #295, opened 2023-05-26
feat(devspace): remove resource requests and leave limits unaltered

kaldorn:
We still want to leave limits on a replaced pod. According to the Devspace docs this should be the correct way to do it for _all_ containers, should there happen to be more than one.

Source: https://www.devspace.sh/docs/configuration/dev/modifications/patches#custom-patches

## Jira ID

[XX-XX]

### DIFF
--- a/templates/devspace.yaml.tpl
+++ b/templates/devspace.yaml.tpl
@@ -161,6 +161,8 @@ dev:
       - op: remove
         path: spec.containers[0].securityContext
       - op: remove
+        path: spec.containers.*.resources.requests
+      - op: remove
         path: spec.containers[0].livenessProbe
       - op: remove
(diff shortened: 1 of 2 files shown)


templates/devspace.yaml.tpl
```diff
@@ -160,6 +160,8 @@ dev:
         value: Always
       - op: remove
         path: spec.containers[0].securityContext
+      - op: remove
```
jaredallard, line 163: This seems useful, but _technically_ `resourcer` is already doing this today, since it mutates the pod before it is created.

kaldorn: Ah. I did not consider that. We can close this then.

PR #297, opened 2023-06-05
fix(devspace): do not automatically forward the DLV port

kaldorn:
When launching two services with devspace port-forwarding the same DLV port causes a conflict. Removing this port-forward is one way to fix it. 

Ideally, we should add functionality to _only_ port-forward the DLV port and add an optional flag to specify the port that then gets set it the vscode json for DLV debugging for multiple services.

## Jira ID

[XX-XX]


templates/devspace.yaml.tpl
```diff
@@ -77,7 +77,9 @@ vars:
   BOX_REPOSITORY_URL: $(yq -r '.storageURL' "$HOME/.outreach/.config/box/box.yaml")
 
 
-  DLV_PORT: 42097
+  DLV_PORT:
+    source: env
+    default: 42097
```
jkinkead, line 82: Does it Just Work to set a port via an environment variable? Like, is it communicated to the running application correctly?

I see it plumbed through in a few places, and it would be great if this more or less worked.

jkinkead (approved):
Can you open a ticket for any remaining work to plumb this through a flag to devenv (or wherever)?

templates/devspace.yaml.tpl
kaldorn: Yes. That is entirely my understanding of it.

daniel-dostal84 (approved):

PR #305, opened 2023-06-12
feat: add automemlimit service activity on by default

jaredallard:
Adds a `automemlimit` service activity that sets `GOMEMLIMIT`[1] by
default to be 90% of the available memory for the cgroup that the
current Go process is running under. This enables the garbage collector
to better be aware of how much memory the cgroup has and to take action
when getting close to it. This will prevent OOMs at the expense of
possibly more CPU cycles.

Also updated `gobox` to get a non-revert version of the new tracing
functionality in.

[1]: https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications


templates/cmd/main.go.tpl
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/getoutreach/gobox/pkg/log"
 	"github.com/getoutreach/gobox/pkg/events"
 	"github.com/getoutreach/gobox/pkg/trace"
+	"github.com/getoutreach/stencil-golang/pkg/serviceactivities/automemlimit"
```
deregtd, line 22: uh, this doesn't exist.  you forget to `git add` some stuff?

jaredallard: probably

deregtd (approved):
90% magic, 10% factz

PR #306, opened 2023-06-14
feat: Add basic binary sync mode for devspace

marekfexa-outreach:
Add support for binary sync in devspace mode. This way we don't need to make dev machine out of devspace pod, but just synchronize binaries and have file system watcher on dev-slim package.

## Jira ID

[QF-880]

## Notes for your reviewers

[QF-880]: https://outreach-io.atlassian.net/browse/QF-880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
+++ b/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
@@ -390,15 +390,6 @@ profiles:
             disableDownload: true
             waitInitialSync: true
       - op: add
-        path: dev.app.patches
-        value: 
-          op: replace
-          path: spec.containers[0].resources
-          value:
-            limits:
-              cpu: 200m
-              memory: 400Mi
-      - op: add
         path: hooks
         value:
           name: reset-dev
--- a/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
+++ b/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
@@ -393,10 +393,11 @@ profiles:
         path: dev.app.patches
         value: 
           op: replace
-          path: spec.containers[0].resources.limits
+          path: spec.containers[0].resources
           value:
-            cpu: 200m
-            memory: 400Mi
+            limits:
+              cpu: 200m
+              memory: 400Mi
       - op: add
         path: hooks
         value:
--- a/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
+++ b/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
@@ -319,7 +319,6 @@ profiles:
     activation:
       - vars:
           DEVENV_DEV_SKIP_PORTFORWARDING: "true"
-          DEVENV_SYNC_BINARIES: "false"
     patches:
       - op: replace
         path: dev.app.ports
@@ -397,7 +396,7 @@ profiles:
           path: spec.containers[0].resources.limits
           value:
             cpu: 200m
-            memory: 200Mi
+            memory: 400Mi
       - op: add
         path: hooks
         value:
--- a/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
+++ b/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
@@ -413,7 +413,7 @@ profiles:
             disableReplace: true
             workDir: ${DEV_CONTAINER_WORKDIR}
             command: |-
-              ../entrypoint.sh
+              entrypoint
 
   # App Profiles
   # Profiles starting with deployment__ are treated specially by devenv.
--- a/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
+++ b/templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
@@ -24,6 +24,12 @@ vars:
   DEVENV_DEPLOY_VERSION:
     source: env
     default: "latest"
+  # DEVENV_SYNC_BINARIES:
+  # false: Devspace is build machine that synchronizes source code and runs it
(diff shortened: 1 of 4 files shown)


templates/.vscode/launch.json.tpl
```diff
@@ -54,6 +54,22 @@
         }
       ],
     },
+    {
+      "name": "Attach to dev container (in binary mode)",
```
marekfexa-outreach, line 58: Here we don't need _substitutePath_ magic. I haven't found better way than having another debug configuration

templates/devspace.yaml.tpl
```diff
@@ -387,6 +397,48 @@ profiles:
...
+            disableDownload: true
+            waitInitialSync: true
+      - op: add
+        path: dev.app.patches
+        value: 
+          op: replace
+          path: spec.containers[0].resources.limits
+          value:
+            cpu: 200m
+            memory: 400Mi
```
marekfexa-outreach, line 423: Services have typically just 100Mi for container. Delve needs more than 200Mi.

templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
```diff
@@ -363,6 +373,48 @@ profiles:
...
+            disableDownload: true
+            waitInitialSync: true
+      - op: add
+        path: dev.app.patches
+        value: 
+          op: replace
+          path: spec.containers[0].resources.limits
+          value:
+            cpu: 200m
+            memory: 400Mi
```
jaredallard, line 399: We might just want to make this the default request instead, since that's really the only time we want to replace these (I imagine we want to leave the default limits for services that HAVE bumped them up)

marekfexa-outreach: Will it be enough for devspace build mode? It feels it needs gigabites to do all the building / linting stuff. That why I would keep it different for bin mode, where 400MI is enough and out of this PR

jaredallard: This limit applies to binary sync, right? I'm more thinking that we need to change the default values so they persist because if we don't we're going to limit all of these applications to 200m/400Mi _while_ they are running (since we need it for the build, but technically after build we want to run with whatever limits they had specified, which seems harder to implement). Lemme know if that's not clear!

marekfexa-outreach: Currently, devspace compile mode does not alter resource limits so it goes with what the service has in production. I raised [concern](https://outreach-hq.slack.com/archives/C04M2RNHY93/p1686809080876359) about that. Originally that resource limit was removed for devspace and it was eating significant resources and making devenv starting for resources. 
Devspace in bin mode (this PR) is changing default resource limits to 200m/400Mi to be able to work with its use cases -> delve, hot reloading.

marekfexa-outreach: We can either:

- Remove limit for devspace compile mode - devspace will work, other pods might be struggling
- Create some limit for devspace compile mode - Like 2CPU and 2GB ram. Maybe more.

I would rather keep that change out of this PR, since I'm touching this part just for devspace binary mode.

marekfexa-outreach: I changed these lines to overwrite whole resources section. Previous approach was not working when some teams defined `resource.request` for devenv that were above the current limits for binary sync. I'm still thinking for better solution, but still I think this is probably good enough for now.

templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
```diff
@@ -363,6 +373,48 @@ profiles:
...
+            "${DEVENV_DEVSPACE_BIN}" reset pods -s
+    merge: 
+      dev:
+        app:
+          # https://www.devspace.sh/docs/configuration/dev/#dev-terminal
+          terminal:
+            enabled: true
+            disableReplace: true
+            workDir: ${DEV_CONTAINER_WORKDIR}
+            command: |-
```
kaldorn, line 415: My understanding is that this profile will created a devspace pod, copy the local binary over, and then leave a terminal running for engineers to interact with/run their service/debug. Is that the intended goal? If so, we should be abundantly cautious around setting strict limits that will apply to 100% of services.

```
      - op: add
        path: dev.app.patches
        value: 
          op: replace
          path: spec.containers[0].resources.limits
          value:
            cpu: 200m
            memory: 400Mi
```

The CPU is compressible, so I'm less concerned. It might just cause the pod to take longer or probes to timeout. I'm concerned that some binaries will _always_ take more than 400Mi when the service is running and OOMKill and our support load will increase.

marekfexa-outreach: Yeah, I agree. However how to ensure, that we have at least 200Mi of memory? Probably somehow parse the limit and replace it when its lower than 200MB. I'll try that.

marekfexa-outreach: which will be challenging thanks to that lovely "Mi"/"Gi" suffixes

kaldorn: I'm in favor of leaving limits up to teams.

templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
marekfexa-outreach: This part is removed @jaredallard

templates/.snapshots/TestDevspaceYaml-devspace.yaml.tpl-devspace.yaml.snapshot
marekfexa-outreach: Removed this part

jaredallard (approved):

PR #316, opened 2023-07-03
refactor(app jsonnet): add non-dev objects block

kaldorn:
Adds a nonDevelopmentObjects block to clean this up a little bit.
Inside of that block there are two conditional objects: HPA and ArgoRollouts configuration.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/deployments/appname/app.jsonnet.tpl
+++ b/templates/deployments/appname/app.jsonnet.tpl
@@ -312,6 +312,7 @@ local all = {
 // nonDevelopmentObjects defines objects for staging/production environments.
 // Note: The vault secrets here are not related to the development vault secrets operator.
 local nonDevelopmentObjects = {
+  {{- if stencil.Arg "vaultSecrets" }}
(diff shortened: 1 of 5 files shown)


yj-yan (approved):

templates/deployments/appname/app.jsonnet.tpl
```diff
@@ -373,30 +309,101 @@ local all = {
...
-  },
-  production: {
-    minReplicas: {{ (stencil.Arg "hpa.env.production.minReplicas") }},
-    maxReplicas: {{ (stencil.Arg "hpa.env.production.maxReplicas") }},
-  },
-};
+// nonDevelopmentObjects defines objects for staging/production environments.
+// Note: The vault secrets here are not related to the development vault secrets operator.
+local nonDevelopmentObjects = {
+  // VaultSecrets to be deployed
```
jaredallard, line 315: Add an `if` here to only render the comment when there is data?

PR #322, opened 2023-07-17
fix(monitoring): remove memory alerts

jaredallard:
I don't believe in memory alerts, to begin with. They also don't work with the new automemlimit work which encourages Go to use 90% of available RAM.


kaldorn (approved):
Noting here that the pod restarts alert will catch OOMKills and is more useful.

PR #325, opened 2023-07-18
fix: Upper bound for datadog terraform provider

joeysmithoutreach:

This PR puts an upper bound on the datadog terraform provider version. It looks like a recent version broke something and this prevents us from using that version.

## Jira ID

[DT-3874]

## Notes for your reviewers

[DT-3874]: https://outreach-io.atlassian.net/browse/DT-3874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/monitoring/versions.tf.tpl
+++ b/templates/monitoring/versions.tf.tpl
@@ -5,7 +5,7 @@ terraform {
   required_providers {
     datadog = {
       source  = "datadog/datadog"
-      version = ">= 2.20.0, < 3.0.0"
+      version = ">= 3.0.0, < 3.27.0"
     }
     vault = {
       source = "hashicorp/vault"
--- a/templates/monitoring/versions.tf.tpl
+++ b/templates/monitoring/versions.tf.tpl
@@ -5,7 +5,7 @@ terraform {
   required_providers {
     datadog = {
       source  = "datadog/datadog"
-      version = ">= 2.20.0"
+      version = ">= 2.20.0, < 3.0.0"
     }
     vault = {
       source = "hashicorp/vault"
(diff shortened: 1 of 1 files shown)


kaldorn (changes requested):
I don't think we actually want to have the upper-bound limit be [the latest version of the datadog-provider](https://github.com/DataDog/terraform-provider-datadog/releases).

The version that implements what is causing the error for that team, based on the ticket, is https://github.com/DataDog/terraform-provider-datadog/releases/tag/v3.21.0

So we want the upper bounds to likely be 3.20.0.

jaredallard:
@joeysmithoutreach Can we also update the title to be more reflective of what this PR is doing? Thank you!!

templates/monitoring/versions.tf.tpl
```diff
@@ -5,7 +5,7 @@ terraform {
   required_providers {
     datadog = {
       source  = "datadog/datadog"
-      version = ">= 2.20.0"
+      version = ">= 3.0.0, < 3.27.0"
```
jkinkead, line 8: Just clarifying what Nate said: We saw issues with a specific version. If `3.27.0` works fine (if Datadog fixed their provider issue), then you should just pin to a major version range:

```suggestion
      version = ">= 3.0.0, < 4.0.0"
```
If you didn't confirm it worked on this version, you should pin to the latest that works.

joeysmithoutreach: Looks like they did fix that issue, so updating to use anything between 3 and 4 (exclusive)

kaldorn (approved):

PR #326, opened 2023-07-20
fix: Create PV + PVC for asdf shims.

jkinkead:
asdf doesn't have a good way to detect missing shims, so keeping this matching installs will make things work better.

---

Without this, every new devspace pod requires a reshim run.

### DIFF
--- a/templates/devspace.yaml.tpl
+++ b/templates/devspace.yaml.tpl
@@ -239,6 +239,19 @@ dev:
           mountPath: /home/dev/.asdf/installs
           name: devspace-asdfcache
 
+      # asdf shims.
+      - op: add
+        path: spec.volumes
+        value:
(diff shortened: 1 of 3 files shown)


templates/devspace.yaml.tpl
```diff
@@ -239,6 +239,19 @@ dev:
           mountPath: /home/dev/.asdf/installs
           name: devspace-asdfcache
 
+      # asdf shims.
+      - op: add
```
jaredallard, line 243: Why not reused the `asdfcache` PVC and instead use [`subPath`](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) on each of these?

jaredallard:
Worked locally :)

jaredallard (approved):

PR #341, opened 2023-08-08
fix(template): increase resources for local env in template

honzap:

After creating a new service it's not possible to build/run it on local environment because of limited resources. This should be then enough for local environment to work without problems (and not failing builds).


honzap:
It should not be in the template. We need to find a different way.

honzap:
https://outreach-io.atlassian.net/browse/DT-3950

PR #343, opened 2023-08-08
fix: devenv app run port forwarding

joeysmithoutreach:

Change devenv app run flags to handle 3 states:
(default) forward only the delve port
--forward-all-ports (-f): do that
--skip-portforwarding (-p): don't forward any ports.

See also: https://github.com/getoutreach/devenv/pull/749

## Jira ID

[DT-3900]

## Notes for your reviewers

[DT-3900]: https://outreach-io.atlassian.net/browse/DT-3900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/devspace.yaml.tpl
```diff
@@ -51,6 +51,9 @@ vars:
   DEVENV_DEV_SKIP_PORTFORWARDING:
     source: env
     default: "false"
+  DEVENV_DEV_ONLY_FOWARD_DLV:
+    source: env
+    default: "false"
```
malept, line 56: thought: For some reason, I thought the current default was `true`.

templates/devspace.yaml.tpl
```diff
@@ -51,6 +51,9 @@ vars:
   DEVENV_DEV_SKIP_PORTFORWARDING:
     source: env
     default: "false"
+  DEVENV_DEV_ONLY_FOWARD_DLV:
```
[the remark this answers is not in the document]
joeysmithoutreach: Dang. Good catch with the typo. I was misspelling forward all over the place, and I thought I fixed them all.

templates/devspace.yaml.tpl
```diff
@@ -332,6 +335,17 @@ profiles:
             labelSelector: 
               app: ${DEVENV_DEPLOY_APPNAME}
 
+  - name: onlyForwardDlv
```
yj-yan, line 338: Should another profile for ForwardAllPorts mode be added?

templates/devspace.yaml.tpl
jaredallard: I believe the intent is that the devenv PR sets it to true, which seems reasonable to me.

jaredallard (approved):
Seems like this adds just a forward deleve only profile, which seems like it makes sense.

templates/devspace.yaml.tpl
joeysmithoutreach: That is correct.

templates/devspace.yaml.tpl
joeysmithoutreach: From the perspective of devspace yaml the "default" is to forward all ports. However, by default the devenv code sets DEVENV_DEV_ONLY_FORWARD_DELVE=true, so from the end user perspective, only forwarding the delve port is the default.

I think we can avoid another profile because of that

PR #346, opened 2023-08-10
fix(DT-3954): Fixing shutdown handling logic for graceful/unexpected shutdowns

deregtd:
* Made a new SignalError to allow wrapping a shutdown initiated by a specific OS signal
* Took the (now-ballooning) logic for handling the different shutdown situations out of the service templates and encapsulated it into a HandleShutdownConditions function in stencil-golang, simplifying the templated code that goes into main.go for services


pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -38,18 +40,24 @@ func New() *ServiceActivity {
...
 // Run runs the shutdown service activity
 func (s *ServiceActivity) Run(ctx context.Context) error {
-	// listen for interrupts and gracefully shutdown server
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	// listen for interrupt, terminated, and hangup signals and gracefully shutdown server
+	cInt := make(chan os.Signal, 1)
+	signal.Notify(cInt, os.Interrupt)
+	cTerm := make(chan os.Signal, 1)
+	signal.Notify(cTerm, syscall.SIGTERM)
```
jaredallard, line 47: Use `os.Kill` instead? More portable.

jaredallard: Not that the program is going to really be able to handle this anyways...

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -38,18 +40,24 @@ func New() *ServiceActivity {
 
 // Run runs the shutdown service activity
 func (s *ServiceActivity) Run(ctx context.Context) error {
-	// listen for interrupts and gracefully shutdown server
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	// listen for interrupt, terminated, and hangup signals and gracefully shutdown server
```
jaredallard, line 43: Can't you cant the mono channel and check what the type of the signal is? It seems better than three signal.Notify calls+channels.

pkg/serviceactivities/shutdown/shutdownsignal.go
```diff
@@ -0,0 +1,42 @@
...
+		return "Terminated"
+	case SignalHangUp:
+		return "HangUp"
+	}
+	return fmt.Sprintf("Error: %d", s)
+}
+
+// List of signals for thue ShutdownFromSignalError
+const (
+	// SignalInterrupt is SIGINT
```
jaredallard, line 26: I'm not a huge fan of duplicating this since this is sorta just a copy of the `signal` package.  Can't we just take `os.Signal` as the `Signal` value here?

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -62,3 +70,23 @@ func (s *ServiceActivity) Close(_ context.Context) error {
 	close(s.done)
 	return nil
 }
+
+// HandleShutdownConditions encapsulates the shutdown logging logic for services into a simple function, and returns
+// a new error code to set, if any
+func HandleShutdownConditions(ctx context.Context, err error) *int {
```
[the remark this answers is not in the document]
deregtd: I was trying to keep the current logic similar.  I'm not sure if people are doing horrible things with exit codes elsewhere or if we can treat exitcode as entirely stencil-golang-owned.

templates/cmd/main.go.tpl
```diff
@@ -158,11 +158,9 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
 {{ file.Block "startup" }}
 	// <</Stencil::Block>>
 
-	if err := async.RunGroup(acts).Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
-		log.Error(ctx, "shutting down service", events.NewErrorInfo(err))
-		return
+	err = async.RunGroup(acts).Run(ctx)
+	updateCode := shutdown.HandleShutdownConditions(ctx, err)
+	if updateCode != nil {
```
jaredallard, line 163: Ah, that's why you're doing `*int`. I guess :D

pkg/serviceactivities/shutdown/shutdown.go
deregtd: we currently don't trap kill, and i think we can't, because, like you say, that's the end of everything immediately, right?

pkg/serviceactivities/shutdown/shutdown.go
deregtd: Different os implementations handle what it sends back differently, unfortunately.  It didn't look like something I could trust. :(

pkg/serviceactivities/shutdown/shutdownsignal.go
deregtd: All of the os structures are platform-specific, so it feels bad to rely on things inside the OS-specific files. If you look into it in VSC, it goes into os_posix.go for example.

pkg/serviceactivities/shutdown/shutdown.go
jaredallard: yeah, gdi, i conflated them both again. Ignore me. 😢

pkg/serviceactivities/shutdown/shutdown.go
jaredallard: Really? I'd expect if we pass `os.Interrupt`, `syscall.SIGTERM`, and `syscall.SIGHUP` we'd only every get those values... maybe I'm wrong though?

pkg/serviceactivities/shutdown/shutdownsignal.go
jaredallard: Totally, but I don't _think_ we're at the point where this matters that much. Right? We're already hard depending on those specific syscalls right now, anyways.

I could see the argument for an interface ontop of it later, if we started doing more complicated things.

deregtd: we can try it for now and see if we get compile errors somewhere in the system

deregtd: actually we can't because os.signal only has term and kill

deregtd: oh i see what you're proposing.  yeah the strings will be assy for other signals but probably good enough, lemme see if that works e2e.

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -7,15 +7,28 @@ package shutdown
...
 
+// FromSignalError is an error struct used by the Shutdown activity to indicate which signal caused the shutdown.
+type FromSignalError struct {
+	Signal os.Signal
+}
+
+// Error satisfies the error interface
+func (s FromSignalError) Error() string {
+	return fmt.Sprintf("shutting down due to interrupt: %v", s.Signal)
+}
```
jaredallard, line 30: nit: I'd prefer `SignalError` as the name.

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -38,18 +51,16 @@ func New() *ServiceActivity {
...
 	select {
-	case out := <-c:
+	case sig := <-c:
 		// Allow interrupt signals to be caught again in worse-case scenario
 		// situations when the service hangs during a graceful shutdown.
 		signal.Reset(os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
-
-		err := fmt.Errorf("shutting down due to interrupt: %v", out)
-		return orerr.ShutdownError{Err: err}
+		return orerr.ShutdownError{Err: FromSignalError{Signal: sig}}
```
jaredallard, line 63: nit: Add a `NewSignalError` function? It'll be cleaner if we ever need to add tests someday.

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -62,3 +73,22 @@ func (s *ServiceActivity) Close(_ context.Context) error {
 	close(s.done)
 	return nil
 }
+
+// HandleShutdownConditions encapsulates the shutdown logging logic for services into a simple function, and returns
+// a boolean indicating if it was a graceful shutdown or not
```
jaredallard, line 78: y u no 72 char word wrap. it's the hot new thing

jaredallard, line 78: It's not really _was_

```suggestion
// HandleShutdownConditions encapsulates the shutdown logging logic for
// services into a simple function, and returns a boolean indicating if
// a graceful shutdown has started or not.
```

jaredallard (commented):
Some nitpicks!

pkg/serviceactivities/shutdown/shutdown.go
```diff
@@ -19,14 +19,18 @@ import (
...
 )
 
-// FromSignalError is an error struct used by the Shutdown activity to indicate which signal caused the shutdown.
-type FromSignalError struct {
+// SignalError is an error struct used by the Shutdown activity to indicate which signal caused the shutdown.
+type SignalError struct {
 	Signal os.Signal
 }
 
+func NewSignalError(s os.Signal) SignalError {
```
jaredallard, line 27: you missed a comment here :(

deregtd: eff

jaredallard (approved):
approving because im sure you'll add that comment and i don't want to come back after a single comment to look at this again

deregtd:
@jaredallard it would fail tests (due to lint) anyway :)

PR #348, opened 2023-08-11
fix(gomod): manage pkg sub module, add tests for native extension

jaredallard:
Previously, we've not been updating/managing the `pkg` sub-module in
this repository because it doesn't follow our versioning scheme. So,
because of this, user's could get issues when a change was made that was
breaking in the templates that needed that module to be updated (they
are meant to only ever be used together). Sadly, because of sub module
versioning, they are being tracked as different versions.

To handle this, we now manage the version in our processor. I attempted
to have this always use the latest version (e.g., `main`, like `go mod tidy`
supports) but, that didn't work. So, we got a nice little cleanup of the 
`MergeGoMod` native extension as well as scenario test coverage
of it instead.


yj-yan (approved):
LGTM

PR #349, opened 2023-08-12
feat: support coralogix

zaquestion:
Adds initial support for leveraging the coralogix terraform.
- coralogix api key was added at the listed vault path

A working plan leveraging these changes can be seen here: https://github.com/getoutreach/orgservice/pull/782#issuecomment-1675474276
I also tested by applying the plan locally, and subsequently destroying
it as to keep a clean tfstate.


kaldorn (changes requested):
We appreciate the help/PR. 
DT had discussed putting this in a private repository instead. Would you be against us moving this and closing this PR?

zaquestion:
Per offline discussion with @kaldorn we're gonna move this to `stencil-outreach`

PR #354, opened 2023-08-16
feat(grpcClients): move to Ruby 3

jaredallard:
Now that Outreach's.. special service... has moved to Ruby 3, we can
move grpcClients to use that by default. So, here we are. Ruby 3. Wow!

Also a misc change to ignore PRs to `main` in VSCode, since those are
created during the release process and then closed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.17.0
+  shared: getoutreach/shared@2.18.0
   queue: eddiewebb/queue@1.8.4
 
 parameters:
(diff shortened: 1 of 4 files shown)


malept:
Has this been tested with any repos using this functionality in production? I have no idea if the Ruby templates need to also be updated.

jaredallard:
It has _not_, however my thinking here is that the ruby gems are being used in 3.1.3 _anyways_ in our special service... so I couldn't see how changing this for the go repos could break anything.

I'm not the Ruby expert though, so please let me know if I'm overlooking anything! (cc @captain-enjoyable, who _might_ have thoughts)

malept (approved):
I'm OK with this if @captain-enjoyable is OK with this.

captain-enjoyable (approved):

PR #358, opened 2023-08-17
fix(devspace): explicitly set requests/limits on devspace container

malept:

This should fix problems some people were having with OOM kills in various places by default.

## Jira ID

[DT-3950]

## Notes for your reviewers

Ignore the whitespace changes, those were autogenerated by VSCode.

[DT-3950]: https://outreach-io.atlassian.net/browse/DT-3950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/devspace.yaml.tpl
+++ b/templates/devspace.yaml.tpl
@@ -57,7 +57,7 @@ vars:
   E2E:
     source: env
     default: "false"
-  
+
 
   # devenv passes in paths to binaries it uses (ensuring the supported versions are used)
   # devenv bin that triggered devspace
@@ -94,6 +94,13 @@ vars:
 deployments:
   app:
     namespace: ${DEVENV_DEPLOY_NAMESPACE}
+    resources:
+      requests:
+        cpu: 1
+        memory: 1Gi
+      limits:
+        cpu: 5
+        memory: 10Gi
     # This deployment uses `kubectl` but you can also define `helm` deployments
     kubectl:
       manifests:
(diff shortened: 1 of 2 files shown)


templates/devspace.yaml.tpl
```diff
@@ -94,6 +94,13 @@ vars:
 deployments:
   app:
     namespace: ${DEVENV_DEPLOY_NAMESPACE}
+    resources:
+      requests:
+        cpu: 1
+        memory: 1Gi
```
joeysmithoutreach, line 100: Do we know how this behaves in a regular devenv environment when multiple services use this as their default? The change looks good, but my only worry is that by setting this as a minimum we might limit how many small services people could run at one time.

malept: I think the question becomes, how many devspace containers do we expect folks to have running at once?

joeysmithoutreach (approved):

templates/devspace.yaml.tpl
malept: We talked about it offline - to be clear this only affects deploying apps into the devenv from source (e.g., `devenv apps run`).

templates/devspace.yaml.tpl
```diff
@@ -94,6 +94,13 @@ vars:
...
+      limits:
+        cpu: 5
+        memory: 10Gi
```
malept, line 103: This was a pretty arbitrary set of values, we can change it if it causes problems.

PR #363, opened 2023-09-05
fix: don't swallow panics

alexandrkara-outreach:

In https://github.com/getoutreach/stencil-golang/pull/121/files a change was made to force exit the code while leaving main(), which was also executed when a panic was being thrown, swallowing and "hiding" it.

This change will at least print it to stderr for debugging.

Not sure whether we should instead make a formatted log entry but this is how panics used to be surfaced before.

## Jira ID

[XX-XX]

## Notes for your reviewers

TESTED=not tested

### DIFF
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -73,7 +73,7 @@ func main() { //nolint: funlen // Why: We can't dwindle this down anymore withou
   exitCode := 1
 	defer func() {
 		if r := recover(); r != nil {
-			fmt.Fprintf(os.Stderr, "Recovered from panic: %+v\n%s\n", r, string(debug.Stack()))
+			panic(r)
 		}
 		os.Exit(exitCode)
 	}()
--- a/templates/cmd/main.go.tpl
+++ b/templates/cmd/main.go.tpl
@@ -72,6 +72,9 @@ type dependencies struct{
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
   exitCode := 1
 	defer func() {
+		if r := recover(); r != nil {
+			fmt.Fprintf(os.Stderr, "Recovered from panic: %+v\n%s\n", r, string(debug.Stack()))
+		}
 		os.Exit(exitCode)
 	}()
 


templates/cmd/main.go.tpl
```diff
@@ -72,6 +72,9 @@ type dependencies struct{
 func main() { //nolint: funlen // Why: We can't dwindle this down anymore without adding complexity.
   exitCode := 1
 	defer func() {
+		if r := recover(); r != nil {
+			fmt.Fprintf(os.Stderr, "Recovered from panic: %+v\n%s\n", r, string(debug.Stack()))
+		}
```
jaredallard, line 77: Sigh. I always forget that a `defer` will just swallow a panic. Good catch, thank you!

This seems like it's a bit lossy compared to the original panic:

**This**: https://go.dev/play/p/JGlQ4lDZv5t?v=goprev
**Original**: https://go.dev/play/p/Qi2xCxybf8R?v=goprev

I suggest we just panic again ([go playground [link]](https://go.dev/play/p/vDS_BRsWoHz?v=goprev)):

```suggestion
		if r := recover(); r != nil {
			panic(r)
		}
```

This is generally what is done when you want to re-throw a panic. It introduces another layer, but I think it's better to not hide the fact that we're handling and re-throwing it.

alexandrkara-outreach:
OK, yeah, that makes sense. On the other hand, it swallows the exit code, if we wanted to set it to a specific value.

jaredallard (approved):

alexandrkara-outreach:
... and prevents us from formatting the panic with `log.Error()`. But any surfacing of the panic is better than just hiding it.

jaredallard:
> OK, yeah, that makes sense. On the other hand, it swallows the exit code, if we wanted to set it to a specific value.

That's true, the intent of the `exitCode` functionality was never to replace's the go standard behaviour though. So, I'm OK with it (we also don't expose that being set by users of stencil-golang today, so I feel further confident in that decision).

jaredallard:
> ... and prevents us from formatting the panic with `log.Error()`. But any surfacing of the panic is better than just hiding it.

Given we haven't been formatting panic logs (at least, I don't think we have been?), I think that serializing the panics and logging them is out of scope of this. If/when we started logging/formatting them, then yes! Totally agree this should be changed.

zaquestion:
We generally try to format them tho...

https://github.com/getoutreach/gobox/blob/ddeb84b00e925e09babf333c73b02239da1d2682/pkg/cli/shutdown.go#L34C1-L40

https://github.com/getoutreach/services/blob/f7ff2cfdf4ff9615a1808fad4b84f0ca0d65cece/pkg/grpcx/server.go#L313-L332

So I definitely wish we'd at least tried to take that approach here... Might as well have just removed the defer otherwise

PR #361, opened 2023-09-04
fix(node): upgrade build nodejs version

jaredallard:
Updates then nodejs version used by build tooling
to be v18.17. This is NOT used for gRPC generation.

PR #364, opened 2023-09-05
test: update tool-versions snapshots

malept:

Follow-up to #361, fixes the CI failure.


yj-yan (approved):
Upgrade node version, LGTM

PR #369, opened 2023-09-08
fix(rpc): indent properly when no gRPC opts/init

jaredallard:
Fixes indentation, whitespace, and output when there are no gRPC opts or
init functions. Otherwise the linter could yell.

### DIFF
--- a/templates/internal/appName/rpc/rpc.go.tpl
+++ b/templates/internal/appName/rpc/rpc.go.tpl
@@ -158,18 +158,18 @@ func (gs *GRPCService) StartServers(ctx context.Context, servers *Servers, opts.
 
 		{{- end }}
 	// end gRPC server option initialization injected by modules
-	{{- end }}
 
+	{{- end }}
+  {{- $grpcServerOptions := stencil.GetModuleHook "internal/rpc/grpcServerOptions" }}
+  {{- if $grpcServerOptions }}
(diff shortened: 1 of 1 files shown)


templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -158,18 +158,18 @@ func (gs *GRPCService) StartServers(ctx context.Context, servers *Servers, opts.
 
 		{{- end }}
 	// end gRPC server option initialization injected by modules
-	{{- end }}
 
+	{{- end }}
+  {{- $grpcServerOptions := stencil.GetModuleHook "internal/rpc/grpcServerOptions" }}
+  {{- if $grpcServerOptions }}
```
jaredallard, line 164: Moved up so we don't render and empty `opts = append(empty, opts)` statement.

templates/internal/appName/rpc/rpc.go.tpl
```diff
@@ -158,18 +158,18 @@ func (gs *GRPCService) StartServers(ctx context.Context, servers *Servers, opts.
 
 		{{- end }}
 	// end gRPC server option initialization injected by modules
-	{{- end }}
```
jaredallard, line 161: Whitespace is behind the `{{- end }}` now so we only render it if we had stuff to write.

kaldorn (approved):

PR #370, opened 2023-09-09
fix(golang): upgrade to 1.20.8

jaredallard:


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.18.0
+  shared: getoutreach/shared@dev:first
```
kaldorn, line 5: Should this be dev:first?

kaldorn (approved):
Looks good. One question about the shared orb.

.circleci/config.yml
jaredallard: It wants it to be :( I'll manually set it to `2.19.2` for now

PR #377, opened 2023-10-03
fix(grpc): upgrade Node.js version to 18

malept:

Node.js 16 is EOL as of 2023-09-11 (per the [Node.js Release working group](https://github.com/nodejs/Release)).

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -172,7 +172,7 @@ arguments:
         - number
   versions.grpcClients.nodejs:
     description: Nodejs version to use for gRPC clients
-    default: "16.19.0"
+    default: "18.17.1"
     schema:
       type:
         - string


kaldorn (approved):
Does this empower us to remove 16 from outreach-docker?

manifest.yaml
```diff
@@ -172,7 +172,7 @@ arguments:
         - number
   versions.grpcClients.nodejs:
     description: Nodejs version to use for gRPC clients
-    default: "16.19.0"
+    default: "18.17.1"
```
jaredallard, line 175: I think we're stuck with this until Giraffe upgrades, technically. We could just upgrade it anyways

kaldorn: Is giraffe bootstrapped? I can't recall the difference between Giraffe and Gazelle, and if Gazelle has replaced Giraffe yet.

malept:
> Does this empower us to remove 16 from outreach-docker?

Probably, yeah.

manifest.yaml
malept: > Is giraffe bootstrapped?

No

> if Gazelle has replaced Giraffe yet

Not yet.

PR #386, opened 2023-10-17
fix(node): use Github Packages hosted npm packages

jaredallard:
Moves to use Github Packages hosted npm packages for `@outreach` instead of npm.

As such, this is changed to use the `@getoutreach` scope.

[DT-4053]

[DT-4053]: https://outreach-io.atlassian.net/browse/DT-4053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/_helpers.tpl
```diff
@@ -165,7 +165,7 @@ nodejs:
     version: {{ .version }}
 {{- end }}
   devDependencies:
-  - name: "@outreach/eslint-config"
+  - name: "@getoutreach/eslint-config"
```
jaredallard, line 168: I need to copy this package over first.

templates/_helpers.tpl
```diff
@@ -148,10 +148,10 @@ nodejs:
     version: ^1.3.5
   - name: "@grpc/proto-loader"
     version: ^0.5.5
-  - name: "@outreach/grpc-client"
+  - name: "@getoutreach/grpc-client"
```
jaredallard, line 151: Small chance I need to update `devbase` to support `@getoutreach` packages.

malept (approved):

PR #390, opened 2023-10-19
fix: Add a 'team' tag to the main Datadog dashboard

lelandbatey:

Right now we don't tag the main `Terraform: {servicename}` datadog dashboard with the team that owns the service, even though we do add a link to that team in the description of the service.

Since we're planning to use "does the dashboard have a team" as a metric for whether or not to clean up that dashboard, these Terraform dashboards (which we use a lot) should have the team added.

## Jira ID

There's no Jira for this; I saw that this is missing and I'm fixing it now.

### DIFF
--- a/templates/monitoring/dashboard.tf.tpl
+++ b/templates/monitoring/dashboard.tf.tpl
@@ -373,6 +373,7 @@ module "temporaldashboard_otel" {
   source      = "git@github.com:getoutreach/monitoring-terraform.git//modules/dd-dashboards"
   name        = "Terraform: {{ .Config.Name | title }} Temporal OTEL"
   description = "Managed by terraform in github.com/getoutreach/{{ .Config.Name }}"
+  tags        = ["team:{{ stencil.Arg "reportingTeam" }}"]
 
   parameters = local.dashboard_parameters
   sections = [


templates/monitoring/dashboard.tf.tpl
```diff
@@ -373,6 +373,7 @@ module "temporaldashboard_otel" {
   source      = "git@github.com:getoutreach/monitoring-terraform.git//modules/dd-dashboards"
   name        = "Terraform: {{ .Config.Name | title }} Temporal OTEL"
   description = "Managed by terraform in github.com/getoutreach/{{ .Config.Name }}"
+  tags        = ["team:{{ stencilArg "reportingTeam" }}"]
```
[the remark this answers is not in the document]
malept: Arguably the dashboards need to move to stencil-outreach since it references monitoring-terraform (not public)

malept, line 376: suggestion: arg syntax (I only know this because I was changing Stencil templates yesterday)
```suggestion
  tags        = ["team:{{ stencil.Arg "reportingTeam" }}"]
```

malept (changes requested):
This is probably a fine short term fix but per my comment, the dashboards should move to stencil-outreach and then we can use the suggestion by Zaq.

templates/monitoring/dashboard.tf.tpl
lelandbatey: Thank you on catching the syntax error!

lelandbatey:
I was operating off of an old `main`; this does belong in stencil-outreach, so I've moved it to that repo: https://github.com/getoutreach/stencil-outreach/pull/252

PR #398, opened 2023-11-03
feat(node): upgrade to v20

jaredallard:
`20` is now LTS, and since we use it for node clients, we will also use
it for build tooling.


malept (commented):
This is slightly different than the Node.js services in use, so we might want to sync the minor versions so we can utilize the default versions installed instead of devs/CI potentially needing to install extra versions.

PR #402, opened 2023-11-10
feat: Improve dockerfile step

voslartomas:

This change will make docker build to cache mod download step, until there is really something changed in dependencies.

https://docs.docker.com/build/guide/mounts/

## Jira ID

[XX-XX]

### DIFF
--- a/templates/deployments/appname/Dockerfile.tpl
+++ b/templates/deployments/appname/Dockerfile.tpl
@@ -9,15 +9,18 @@ ENV GOPRIVATE github.com/{{ .Runtime.Box.Org }}/*
 ENV CGO_ENABLED {{ stencil.ApplyTemplate "cgoEnabled" | trim }}
 WORKDIR /src
 
-# Copy our source code into the container for building
-COPY . .
-
 ## <<Stencil::Block(beforeBuild)>>
 {{ file.Block "beforeBuild" }}
 ## <</Stencil::Block>>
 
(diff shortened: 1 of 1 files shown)


templates/deployments/appname/Dockerfile.tpl
```diff
@@ -9,15 +9,18 @@ ENV GOPRIVATE github.com/{{ .Runtime.Box.Org }}/*
...
-
 ## <<Stencil::Block(beforeBuild)>>
 {{ file.Block "beforeBuild" }}
 ## <</Stencil::Block>>
 
 # Cache dependencies across builds
-RUN --mount=type=ssh --mount=type=cache,target=/go/pkg go mod download
+RUN --mount=type=ssh --mount=type=cache,target=/go/pkg \
+    --mount=type=bind,source=go.sum,target=go.sum \
+    --mount=type=bind,source=go.mod,target=go.mod \
```
jaredallard, line 19: This will break for repos that have `go.work` and multiple go modules (this is currently used by some repos). We might want to adapt it for that, but I don't know how we would right now. 

Does `go mod download` actually do anything if the modules already exist on disk (in the cache mount?)?

voslartomas: I guess it should just take it from the cache, I guess currently it is only problem because there is still no DLC enabled, once that gets propagated it should be fine. So we can close this.

PR #405, opened 2023-11-15
fix(rubyClient): upgrade dependencies, unmanage Gemfile.lock

jaredallard:
Updates the dependencies used by the ruby client. Makes `Gemfile.lock`
as static so that we no longer manage it.

### DIFF
--- a/templates/api/clients/ruby/Gemfile.lock.tpl
+++ b/templates/api/clients/ruby/Gemfile.lock.tpl
@@ -1,27 +1,29 @@
 {{- $_ := stencil.ApplyTemplate "skipGrpcClient" "ruby" -}}
+{{ file.Static }}
 PATH
   remote: .
   specs:
-    {{ .Config.Name }}_client (1.63.0)
-      grpc (~> 1.38)
+    {{ .Config.Name }}_client (0.0.0)
+      grpc (~> 1.59)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    google-protobuf (3.19.1)
-    googleapis-common-protos-types (1.3.0)
-      google-protobuf (~> 3.14)
-    grpc (1.42.0)
+    google-protobuf (3.25.0)
+    google-protobuf (3.25.0-arm64-darwin)
+    googleapis-common-protos-types (1.10.0)
       google-protobuf (~> 3.18)
+    grpc (1.59.2)
+      google-protobuf (~> 3.24)
       googleapis-common-protos-types (~> 1.0)
-    rake (13.0.6)
-
+    rake (13.1.0)
 PLATFORMS
   -darwin-20
+  arm64-darwin-22
 
 DEPENDENCIES
   {{ .Config.Name }}_client!
   rake
 
 BUNDLED WITH
-   2.2.29
+   2.3.26
--- a/templates/api/clients/ruby/client.gemspec.tpl
+++ b/templates/api/clients/ruby/client.gemspec.tpl
@@ -11,11 +11,11 @@ Gem::Specification.new do |spec|
   spec.metadata["source_code_uri"] = spec.homepage
   spec.metadata["github_repo"] = "ssh://github.com/getoutreach/{{ .Config.Name }}"
 
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.7.5")
+  spec.required_ruby_version = Gem::Requirement.new(">= {{ stencil.Arg "versions.grpcClients.ruby" }}")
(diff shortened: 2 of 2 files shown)


templates/api/clients/ruby/Gemfile.lock.tpl
```diff
@@ -1,27 +1,29 @@
 {{- $_ := stencil.ApplyTemplate "skipGrpcClient" "ruby" -}}
+{{ file.Static }}
 PATH
   remote: .
   specs:
-    {{ .Config.Name }}_client (1.63.0)
-      grpc (~> 1.38)
+    {{ .Config.Name }}_client (0.0.0)
```
jaredallard, line 6: `0.0.0` is to make it clear that this doesn't coincide with anything.

templates/api/clients/ruby/Gemfile.lock.tpl
```diff
@@ -1,27 +1,29 @@
...
       google-protobuf (~> 3.18)
+    grpc (1.59.2)
+      google-protobuf (~> 3.24)
       googleapis-common-protos-types (~> 1.0)
-    rake (13.0.6)
-
+    rake (13.1.0)
 PLATFORMS
   -darwin-20
+  arm64-darwin-22
```
jaredallard, line 22: I'm not 100% sure if this is good or bad. Maybe @malept would know.

jaredallard:
This fixed the underlying issue (no builds for ruby+arch pairing), so I'm going to go ahead and consider this good unless any of my reviewers reject it :D

templates/api/clients/ruby/Gemfile.lock.tpl
malept: So, if I remember correctly, the suffix number will change every time there's a new macOS release. For the actual function of this section, I'd need to do some research which would happen tomorrow at the earliest.

templates/api/clients/ruby/Gemfile.lock.tpl
```diff
@@ -1,27 +1,29 @@
...
 
 DEPENDENCIES
   {{ .Config.Name }}_client!
   rake
 
 BUNDLED WITH
-   2.2.29
+   2.3.26
```
malept, line 29: thought: I wonder if it would be worth putting the bundler version in the manifest.

templates/api/clients/ruby/client.gemspec.tpl
```diff
@@ -11,11 +11,11 @@ Gem::Specification.new do |spec|
...
   spec.metadata["github_repo"] = "ssh://github.com/getoutreach/{{ .Config.Name }}"
 
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.7.5")
+  spec.required_ruby_version = Gem::Requirement.new(">= {{ stencil.Arg "versions.grpcClients.ruby" }}")
   spec.files         = Dir.chdir(File.expand_path('..', __FILE__)) do
     `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   end
   spec.require_paths = ["lib", "lib/{{ .Config.Name }}_client"]
-  spec.add_dependency 'grpc', '~> 1.38'
+  spec.add_dependency 'grpc', '~> 1.59'
```
malept, line 19: question: should we put this version in the manifest as well since it's referenced in multiple places?

templates/api/clients/ruby/Gemfile.lock.tpl
jaredallard: I honestly don't know, but given we're unmanaging it, I would think once is okay?

templates/api/clients/ruby/client.gemspec.tpl
jaredallard: That sounds like a reasonable improvement

templates/api/clients/ruby/Gemfile.lock.tpl
malept: Oh right, ignore this then

templates/api/clients/ruby/Gemfile.lock.tpl
malept: issue: I think in order for this to work, we'd need to have a post-template rendering hook to run `bundle install` and set the number correctly.

malept: Sure, but that's a valid version. `0.0.0` isn't :disappointed:

malept: Oh, I guess that's an invalid version for new services, sigh.

PR #409, opened 2023-11-27
chore: Promote unstable to rc

yj-yan:
Promote unstable to rc, created by `dt release promote`

**DO NOT MERGE**


malept (changes requested):
I don't really want to upgrade prettier in this release cycle. I deliberately merged the change *after* the RCs went out.

PR #422, opened 2024-01-11
fix: upgrade modules and re-stencil (2024-01-11)

pbr0ck3r:

## Jira ID

[XX-XX]


.circleci/config.yml
```diff
@@ -2,8 +2,8 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.19.2
-  queue: eddiewebb/queue@1.8.4
+  shared: getoutreach/shared@dev:first
```
kaldorn, line 5: This seems not quite right and I can't recall what it should be instead.

kaldorn (changes requested):
One comment/concern, otherwise looks fine.

.circleci/config.yml
pbr0ck3r: Yeah it seemed odd to me too... was having issues with stencilbot so ran stencil manually on this

kaldorn (approved):

.circleci/config.yml
malept: This is because devbase is using unstable. Because of the issues with the dev CircleCI orb (it expires after some time), we may want to use devbase RC / stable instead.

.circleci/config.yml
```diff
@@ -2,8 +2,8 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.19.2
-  queue: eddiewebb/queue@1.8.4
+  shared: getoutreach/shared@2.21.0
```
malept, line 5: suggestion: we may want to use the latest version
```suggestion
  shared: getoutreach/shared@2.24.0
```

pbr0ck3r: Sounds good

malept (approved):

PR #446, opened 2024-03-12
fix: re-add Go 1.22.0 support

malept:

Does what it says on the tin.

This should only be needed for the RC branch.


kaldorn (approved):
Adds .5 to the minor bug fix to bump the major bug fix version.
1.21.5 + 0.0.5 = 1.22.0.

PR #461, opened 2024-03-28
feat: send appName as an identifier in LD context

thezero:
Update LaunchDarkly context by introducing `appName` property in the user context. This property is already used by client-side SDK (OrCA) to identify various apps built in OrCA. In order to have a better overview of usage of flags across our codebase, we're introducing the `appName` property across all LD clients.

This is a backwards compatible change because it doesn't change how existing flags are evaluated. It's not expected to use this new property in feature flag targeting; instead, this will come in handy as a debugging/organization tool.

Recently, there were at least two occasions where this would have been useful:
1. An SDK sending invalid value of user id 
2. An SDK sending incorrect insights data

In both cases, having the `appName` would speed up the fault detection.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/deployments/appname/app.jsonnet.tpl
+++ b/templates/deployments/appname/app.jsonnet.tpl
@@ -169,7 +169,7 @@ local all = {
 				Path: '/run/secrets/outreach.io/launchdarkly/sdk-key',
 			},
 			flagsToAdd: {
-        appName: app.name,
+				appName: app.name,
 				bento: app.bento,
 				channel: if isDev then 'dev' else app.channel,
 			} + if isDev then {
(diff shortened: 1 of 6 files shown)


templates/deployments/appname/app.jsonnet.tpl
```diff
@@ -169,6 +169,7 @@ local all = {
 				Path: '/run/secrets/outreach.io/launchdarkly/sdk-key',
 			},
 			flagsToAdd: {
+        appName: app.name,
```
malept, line 172: suggestion: according to the snapshot test failure, this should be tabs to be consistent with the other jsonnet around it.
```suggestion
				appName: app.name,
```

malept (changes requested):

templates/deployments/appname/app.jsonnet.tpl
malept: But also the tests need to be regenerated (Run `make test` after this change is in, and commit the updated snapshots).

thezero: Thank you, updated (sorry I'm not a golang dev so I'm learning as I go)

malept (approved):

PR #480, opened 2024-04-11
fix(docker): avoid installing curl for the RDS certificates

malept:

In Alpine Linux, `curl` isn't installed by default, but `busybox` is. `busybox` contains a `wget` applet which has the equivalent functionality for what we need here. When the image layer isn't cached, it seems to go from 1.1s to 0.2s.


yj-yan (approved):
Less installment in Docker, nice

PR #489, opened 2024-04-19
fix(goreleaser): correct the name of env block and indents

yj-yan:

It turns out if the cli name contains special character "-", it will breaks the stencil block. Use the index to name the block instead. Also fixed the indent of this file.

## Jira ID

[XX-XX]


malept (changes requested):
I think using index to differentiate blocks is going to make it confusing for users. Can we instead sanitize the CLI name when using it as the block name?

templates/.goreleaser.yml.tpl
```diff
@@ -12,28 +12,29 @@ builds:
...
+    ldflags:
+      - '-w -s -X "github.com/getoutreach/gobox/pkg/app.Version=v{{ "{{" }} .Version {{ "}}" }}"'
+      {{- if not $opts.delibird }}
+      - '-X "main.HoneycombTracingKey={{ "{{" }} .Env.HONEYCOMB_APIKEY {{ "}}" }}"'
+      - '-X "main.TeleforkAPIKey={{ "{{" }} .Env.TELEFORK_APIKEY {{ "}}" }}"'
+      {{- end }}
+    env:
+      - CGO_ENABLED={{ stencil.ApplyTemplate "cgoEnabled" | trim }}
+      {{- $cmd :=  $cmdName | split "-" }}
+      {{- $blockName := (printf "%vAdditionalEnv" $cmd._0) }}
```
malept, line 33: suggestion: this will be problematic if, for example, we made a repo with a bunch of `kubectl` plugins (they are named `kubectl-foo` and `kubectl-bar`). Do underscores work? If so, you can do this:
```suggestion
      {{- $blockName := (printf "%vAdditionalEnv" ($cmdName | replace "-" "_")) }}
```

malept: If underscores don't work, I'd just get rid of it altogether:
```suggestion
      {{- $blockName := (printf "%vAdditionalEnv" ($cmdName | replace "-" "")) }}
```

yj-yan: I see, good point. I will try if underscore works in block.

yj-yan: Underscore does not work in block, replaced "-" with "".

PR #494, opened 2024-04-29
chore: restencil to use unstable devbase instead of a nonexistent branch

malept:

This fixes CI now that https://github.com/getoutreach/devbase/pull/743 is merged.


yj-yan (approved):
I forgot to update this one. Thanks!

PR #497, opened 2024-05-06
fix: upgrade `@getoutreach/eslint-config` to 2.0.0

malept:

Does what it says on the tin.

## Jira ID

[DT-4345]


pbr0ck3r (approved):
Upgrade! Nice work!

PR #506, opened 2024-05-10
fix: allow hyphenated services

MaxMcF:

- This addition ignores a linter error thrown when re-stenciling services with a `_` in their service name. 
- According to the templated comment, we should allow these names:
  - `Why: We allow [-_].` 
- When re-stenciling a `data-sharing-service` (gocode service uses a `_` in the name), the following error is thrown:
  - `api/data-sharing-service/doc.go:8:1: package "data_sharing_service" should not contain - or _ in name (doculint)`  

## Jira ID

[XX-XX]


malept (changes requested):
I'm almost positive that this isn't the only change that needs to be made in order for this to work. Many more of the stencil modules need to be updated in order to handle hyphens.

MaxMcF:
@malept Closing this PR, as it sounds like this will take a significant amount of additional work. Other than this repo and terraform monitor repos, I'm unaware of any other stencil packages that would need this change. If you think this is doable without significant refactoring, please let me know which other stencil repos would require work. Thanks!

PR #513, opened 2024-05-31
fix(nodejs): update @grpc/proto-loader dependency to fix a vulnerability

surya-vl:

In commitbridge, we are importing "github.com/getoutreach/stencil-golang/pkg" and it is adding a dependency of "proto-loader" of verion 0.5.5 and this adding a dependency of "protobufjs" in yarn.lock file with version 6.8.6
Now the issue comes here, there's a vulnerability with this version of the protobufjs and you can [refer this article for more details](https://security.snyk.io/vuln/SNYK-JS-PROTOBUFJS-5756498) 
So I need to use protobufjs versions above 6.11.4 or 7.2.4 and to do that I need to update proto-loader version
I have updated the proto-loader version to 0.7.13 in commitbridge but restencil is degrading it back to 0.5.5 due to the version mentioned in templates/_helpers.tpl file in this repository. So, I would like to update proto-loader version in templates/_helpers.tpl file so that any repository using stencil-golang or stencil-golang/pkg will not use these vulnerability versions of protobufjs.

## Jira ID

[FRI-4247](https://outreach-io.atlassian.net/browse/FRI-4247)

## Notes for your reviewers

[FRI-4247]: https://outreach-io.atlassian.net/browse/FRI-4247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ test: &test
 
 # Branches used for releasing code, pre-release or not
 release_branches: &release_branches # Release branch
-  - release
+  - "release"
   # Pre-releases branch
   - "main"
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.27.0
+  shared: getoutreach/shared@dev:2.27.1-rc.1
   queue: eddiewebb/queue@2.2.1
 
 parameters:
--- a/stencil.lock
+++ b/stencil.lock
@@ -8,7 +8,7 @@ modules:
       version: v0.16.1-rc.1
     - name: github.com/getoutreach/stencil-circleci
       url: https://github.com/getoutreach/stencil-circleci
-      version: v1.13.0
+      version: v1.13.1-rc.1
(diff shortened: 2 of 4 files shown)


malept:
1. Why is the change needed?
2. Why does this PR require a restencil?
3. This change affects every single service which provides a Node.JS gRPC client. This needs to be validated that it won't break anyone.
4. The dependency in question is a pre-1.0 release that is two minor version bumps (it goes from 0.5 to 0.7). What are the changes? Are any of the changes breaking changes? If so, how do they affect our usage of the dependency?

surya-vl:
> 1. Why is the change needed?
> 2. Why does this PR require a restencil?
> 3. This change affects every single service which provides a Node.JS gRPC client. This needs to be validated that it won't break anyone.
> 4. The dependency in question is a pre-1.0 release that is two minor version bumps (it goes from 0.5 to 0.7). What are the changes? Are any of the changes breaking changes? If so, how do they affect our usage of the dependency?

1. I have updated the description with details to answer why this PR is required.
2. This PR doesn't a requirement to do restencil, but I did that to cross-verify that restencil is not degrading the updated versions.

PR #516, opened 2024-06-21
fix(gomod): add toolchain directive

malept:

This should ensure that the Go version specified by the manifest is the same version as the toolchain used.


yj-yan (approved):
Looks good, just need to update the test snapshots.

malept:
> Looks good, just need to update the test snapshots.

This is what happens when I update a repo with the GitHub UI :laughing: :cry:

PR #520, opened 2024-06-27
fix: update all deps

george-e-shaw-iv:
* ran `go get -u ./... && go mod tidy`


malept (approved):
Looks legit

malept:
Any particular reason why we need to do this?

PR #527, opened 2024-07-16
feat(manifest): upgrade Go version to 1.22.5

malept:


yj-yan (approved):
LGTM. I assume the minor version upgrade should be all compatible.

PR #533, opened 2024-08-08
feat(svc): create Run method to encapsulate most of whats in func main

AndrewWinterman:

This PR extracts most of the main function provided by stencil-golang into a
function clients can call instead.

QSS runs a number of peripheral services-- CDC, Ingest, Polaroid, SmartWorker,
etc., each of which need to set up their own main function. To do this, they
basically copy whats in stencil-golang. Wouldn't it be better if we could use a
library function instead? I think it would be, since it eases the burden when
it comes time to update any service-wide behavior.

Of course, we'll still want stencil to provide the kubernetes objects, e.g.
trace.yaml for example, and my hope is we'll stencil in depending on this
method instead of just doing what it does currently inline.

## Jira ID

[XX-XX]


pkg/run/run.go
```diff
@@ -0,0 +1,160 @@
...
+	return s(o)
+}
+
+// WithLogger sets the logger. Otherwise it defaults to olog.New
+func WithLogger(l *slog.Logger) Option {
+	return optionFunc(func(o *runOpts) error {
+		o.log = l
+		return nil
+	})
+}
```
malept, line 58: thought: I wonder if this is the right way to do this, given that `olog`-based logging seems to be package-based now (as opposed to `gobox/pkg/log` which is effectively a global)

pkg/run/run.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2024 Outreach Corporation. All Rights Reserved.
+
+// Description: Entrypoint to run the app
+
+// Package run provides a function that can be invoked inside of main to set up
+// all the service-standard components we expect every service to run.
+//
+// clients should provide any runners they require as part of the app to the
+// Run function with OptAddRunner, and then call `Run` in their main function.
```
malept, line 9: suggestion: it might be worth adding an example of how to use this.

malept (changes requested):
Shouldn't [`templates/cmd/main.go.tpl`](https://github.com/getoutreach/stencil-golang/blob/main/templates/cmd/main.go.tpl) also be changed in this PR?

george-e-shaw-iv (approved):
I've already reviewed the majority of this code. I think Mark has some fair points though

pkg/run/run.go
AndrewWinterman: My thought was that this is more flexible for teams to configure the logger, since some configuration options aren't super accessible yet

AndrewWinterman:
> Shouldn't [`templates/cmd/main.go.tpl`](https://github.com/getoutreach/stencil-golang/blob/main/templates/cmd/main.go.tpl) also be changed in this PR?

I had thought not, I'd like to merge it, use it some test services, and then circle back.

pkg/run/run.go
AndrewWinterman: Added a testable example... should probably add a note that it exists here.

AndrewWinterman:
hit an issue with https://github.com/search?q=org%3Agetoutreach+stencil.AddToModuleHook+%22github.com%2Fgetoutreach%2Fstencil-golang%22+%22serviceActivities%22&type=code

PR #556, opened 2024-11-19
fix(ruby): sync bundler version with upstream

malept:

Upstream uses a newer version of Bundler, which causes conflicts in CI.


ChrisPRobinson (approved):
LGTM

PR #561, opened 2024-12-11
feat: stencil blocks around arch+os in .goreleaser.yml

honzap:

Discussed in [Slack](https://outreach-hq.slack.com/archives/C01CAR6GVPD/p1733906223077049?thread_ts=1733820097.550999&cid=C01CAR6GVPD). Not all services needed both architectures to be built. This makes ability to define it for each of the service individually.

The block set's the default and preserves if there's something else.

It'd be also worth it to think about dropping arm64+darwin as only a few repos/services need assets to be built (for example: clerkcli). 

Tested locally.

## Jira ID

[XX-XX]

### DIFF
--- a/templates/.goreleaser.yml.tpl
+++ b/templates/.goreleaser.yml.tpl
@@ -16,11 +16,27 @@ builds:
     id: &name {{ $cmdName }}
     binary: *name
     goos:
+      {{- $goosBlockName := (printf "%vGoos" ($cmdName | replace "-" "" | replace "_" "")) }}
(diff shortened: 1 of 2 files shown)


honzap:
discussed, DT team will handle it in the [ticket](https://outreach-io.atlassian.net/browse/DT-5311), also related discussion [here](https://outreach-hq.slack.com/archives/C01CAR6GVPD/p1733820097550999) and [here](https://outreach-hq.slack.com/archives/CN9MU7GLW/p1777283734934169).

PR #570, opened 2025-01-14
fix: update stencil-golang/pkg version used by repos

malept:

This version hasn't been updated in a while.

Additionally, sync the Go version in `pkg/go.mod` with the one specified in `go.mod`. This will be synced any time `make fmt` is run (which is also during re-`stencil`-ing)

## Jira ID

[DT-4574]

[DT-4574]: https://outreach-io.atlassian.net/browse/DT-4574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/sync-gomod.sh
```diff
@@ -0,0 +1,42 @@
...
+run_sed() {
+  case "$OSTYPE" in
+  darwin*) # macOS
+    sed -i '' -e "$@"
+    ;;
+  linux*)
+    sed -i -e "$@"
+    ;;
+  esac
+}
```
malept, line 18: Admittedly I did not test this on macOS, but it's very similar to this function:

https://github.com/getoutreach/devbase/blob/1ff9c81010a7e7771d29971daf45dde17f9eaa8c/shell/catalog-sync.sh#L14-L28

ekalinichev: Didn't run the whole thing but verified the mac command

ekalinichev (approved):

PR #577, opened 2025-02-17
fix: move k8s imports from services to k8slib

evadnoob:
Changes services dependencies for k8s libraries over to k8slib


templates/internal/appName/k8s/controller.go.tpl
```diff
@@ -16,10 +16,10 @@ import (
 	"context"
 
 	"github.com/getoutreach/gobox/pkg/log"
-	k8slogging "github.com/getoutreach/services/pkg/k8s/logging"
-	"github.com/getoutreach/services/pkg/k8s/controllers"
-	"github.com/getoutreach/services/pkg/k8s/logging"
-	"github.com/getoutreach/services/pkg/k8s/resources"
+	k8slogging "github.com/getoutreach/k8slib/pkg/k8s/logging"
+	"github.com/getoutreach/k8slib/pkg/k8s/controllers"
+	"github.com/getoutreach/k8slib/pkg/k8s/logging"
```
malept, line 21: suggestion: line 19 and 21 are pointing to the same thing but with multiple names. That should be consolidated.

malept (changes requested):

templates/internal/appName/k8s/controller.go.tpl
malept: Looks like `k8slogging` is unused in the template.
```suggestion
	"github.com/getoutreach/k8slib/pkg/k8s/controllers"
	"github.com/getoutreach/k8slib/pkg/k8s/logging"
```

templates/_helpers.tpl
```diff
@@ -135,7 +135,7 @@ go:
   version: v1.148.6
   # Used for grpcx
 - name: github.com/getoutreach/services
-  version: v1.175.1
+  version: v1.218.0
```
malept, line 138: Updated this version to make sure to get a minimum services version without `services/pkg/k8s/*`.

templates/_helpers.tpl
```diff
@@ -148,6 +148,8 @@ go:
 {{- end }}
 
 {{- if stencil.Arg "kubernetes.groups" }}
+- name: github.com/getoutreach/k8slib
+  version: v1.0.0
```
malept, line 152: question: @evadnoob this (use of `kubernetes.groups` in `service.yaml`) is the only condition in `stencil-golang` where `k8slib` needs to be included, correct?

evadnoob: yes, I think so. I'll update to match the same versions in go.mod for k8slib

malept (approved):

PR #578, opened 2025-02-20
feat: remove KIAM functionality

dvagapov:

- Remove Kiam's code
- Do not touched args `useKIAM` -> because it's still using by some apps

## Jira ID

[XX-XX]

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -45,7 +45,7 @@ arguments:
   aws.useKIAM:
     schema:
       type: boolean
-    default: true
+    default: false
     description: Whether the application uses KIAM for IAM roles or IRSA roles for service accounts.
   oss:
     schema:
--- a/templates/deployments/appname/app.jsonnet.tpl
+++ b/templates/deployments/appname/app.jsonnet.tpl
@@ -231,7 +231,7 @@ local all = {
 						'tollgate.outreach.io/port': '5000',
 						{{- end }}
 						{{- if or (eq "datadog" (stencil.Arg "metrics")) (eq "dual" (stencil.Arg "metrics")) }}
-            datadog_prom_instances_:: [
+						datadog_prom_instances_:: [
 							{
 								prometheus_url: 'http://%%host%%:' +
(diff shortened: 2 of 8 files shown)


manifest.yaml
```diff
@@ -45,7 +45,7 @@ arguments:
   aws.useKIAM:
     schema:
       type: boolean
-    default: true
+    default: false
     description: Whether the application uses KIAM for IAM roles or IRSA roles for service accounts.
```
jan-ludvik, line 49: Should this be removed completely?

templates/deployments/appname/app.jsonnet.tpl
```diff
@@ -61,22 +61,15 @@ local deploymentMetrics = [
...
 			labels+: sharedLabels,
 		},
 	},
 	svc_acct+: ok.ServiceAccount('%s-svc' % app.name, app.namespace) {
 		metadata+: {
 			labels+: sharedLabels,
-			{{- if not (stencil.Arg "aws.useKIAM") }}
 			annotations+: {
 				'eks.amazonaws.com/role-arn': 'arn:aws:iam::{{ .Runtime.Box.AWS.DefaultAccountID }}:role/%s-%s' % [app.bento, app.name]
 			},
```
jan-ludvik, line 72: We could provide terraform role with no permissions by default. We could however run out of role limit in our account. Not sure if we should do that. Maybe we could keep both behind some new parameter in service.yaml - something like `awsRole`?

ekalinichev (approved):

ekalinichev (changes requested):
1. Failed CI
2. Questions from @jan-ludvik aren't answered

templates/deployments/appname/app.jsonnet.tpl
dvagapov: Makes sense to add arg like `awsRole` for custom role, but Dev team also can use override.jsonnet for make this if needed.
I think we can keep here just generic awsRole

ekalinichev (approved):

PR #582, opened 2025-03-04
chore(CONTRIBUTING): add note about snapshots

malept:

Make it explicit that snapshots should be updated as necessary.


malept:
Blocked by #583 as the tests failed (not due to this PR)

ekalinichev (approved):

PR #588, opened 2025-04-14
fix(deployments): disable OTel for local development

richardlarocque:
[DP-6272](https://outreach-io.atlassian.net/browse/DP-6272)

Updates the `app.jsonnet` template so it sends traces directly to
Honeycomb rather than to the collector in local dev mode.

It relies on a special behavior of gobox's pkg/trace that kicks in
when the `CollectorEndpoint` is `""`. [^1]

This is upstreams a fix that appeared earlier [^2].

We prefer to send the traces directly to the collector for
consistency with production.  We run the OTel collector inside the
devenv for this purpose, and so it makes sense to send traces to
the collector in that case (`app.environment == 'development'`).

In local development (`app.environment == 'local_development'`), the
app is running directly on the user's machine, and it won't be able
to talk to devenv.  At least, not unless `devenv tunnel` is up and
running.  Since `make devserver` is intended for local testing, it
seems unreasonable to expect them to actually be running a devenv
and tunnel in this case.

This work-around may cease to work well as we rely on the OTel collector
for more things, but it works well enough for now.  We can always
revisit this when it stops working.  It's also a dev-only change, so it
should be low risk.

[^1]: https://github.com/getoutreach/gobox/pull/476
[^2]: https://github.com/getoutreach/opdp/pull/92/files#r2025781619

[DP-6272]: https://outreach-io.atlassian.net/browse/DP-6272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndrewWinterman (approved):

malept:
FYI, I fixed the PR title since it's used in (public) release notes.

PR #608, opened 2025-08-09
fix: set correct copyright header for OSS repos

malept:

Does what it says on the tin.


ekalinichev (approved):
Good catch!

PR #617, opened 2025-09-30
feat: allow containers to be built for the primary CLI of CLI-only repos

malept:

Previously, if you wanted to do this, you'd have to make your own `Dockerfile` (and sync it yourself every time you restencil with a new Go version) and corresponding infrastructure.

## Jira ID

[DT-4900]

[DT-4900]: https://outreach-io.atlassian.net/browse/DT-4900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/.snapshots/TestRenderDeploymentDockerfile-deployments-appname-Dockerfile.tpl-deployments-testing-Dockerfile.snapshot
```diff
@@ -1,5 +1,5 @@
 (*codegen.File)(# syntax=docker/dockerfile:1.0-experimental
-FROM registry.example.com/foo/golang:1.23.4 as builder
+FROM registry.example.com/foo/golang:1.0.0 as builder
```
malept, line 2: As it turned out, the test was using the wrong config value and so it was grabbing the real golang version instead of the one defined in the test.

ekalinichev (approved):

PR #692, opened 2026-04-02
feat(grpc): migrate Node.js gRPC client templates from eslint to oxlint

malept:

The TL;DR for why we're doing this is because `eslint` and its plugins are causing a lot of strife:
* `eslint` itself is comprised of 99 packages. I didn't even bother to look to see how many the plugins are made of.
* Several of the dependencies have had regular vulnerabilities.
* Other dependencies of eslint and its plugins have been problematic, notably `flatted` as the developer is including non-JS ecosystem files in the npm package (e.g., Go/PHP).

We could upgrade to the latest major version of eslint but it's about the same amount of effort as upgrading to oxlint, and oxlint has much fewer dependencies.

## Jira ID

[DT-5274]

## Notes for your reviewers

- Add .oxlintrc.json.tpl extending @getoutreach/oxlint-config
- Mark .eslintrc.js.tpl for deletion via file.Delete
- Update lint scripts in package.hjson.tpl to use oxlint
- Replace 7 eslint devDependencies with oxlint-config and oxlint

### TODO

- [x] Merge https://github.com/getoutreach/services/pull/763

[DT-5274]: https://outreach-io.atlassian.net/browse/DT-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iamrz1 (approved):
✅ LGTM!

PR #696, opened 2026-04-13
fix: bump nodejs default client version

bogdanlytvynovskyi:
Please read [CONTRIBUTING.md](CONTRIBUTING.md) for additional information on contributing to this repository!

## Jira ID

[XX-XX]

## Notes for your reviewers
Currently installed version of NodeJS on repo level conflicts with the one in generated node client. This is an attempt to get it to latest

---
Rovo Dev code review: <strong>Rovo Dev has reviewed this pull request</strong>
Any suggestions or improvements have been posted as pull request comments.

PR #698, opened 2026-04-14
fix: failing oxlint and test_node_client

PeSme:

Recently oxlint was introduced but it requires new nodejs version, the root-level `.tool-versions` is fine but there is a special `.tool-versions` in `api/clients/node/` which is much older. No idea why we need to have a different version here, but this patch helps. The only occurance of that version is really in https://github.com/getoutreach/stencil-golang/blob/main/templates/api/clients/node/.tool-versions.tpl only.

Without it `stencil` log contains an error

```
error oxlint@1.60.0: The engine "node" is incompatible with this module. Expected version "^20.19.0 || >=22.12.0". Got "20.16.0"
error Found incompatible module.
```

and also the `test_node_client` step fails, normally it's not required so people don't care but we have it required in some services...

## Jira ID

[XX-XX]

## Notes for your reviewers

---
Rovo Dev code review: <strong>Rovo Dev is reviewing this pull request…</strong>
Refresh the page in a few minutes to see the results.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -178,7 +178,7 @@ arguments:
         - number
   versions.grpcClients.nodejs:
     description: Nodejs version to use for gRPC clients
-    default: "20.16.0"
+    default: "22.22.0"
     schema:
       type:
         - string


iamrz1:
Hi, This PR looks like a duplicate of this https://github.com/getoutreach/stencil-golang/pull/696, which was closed in favor of the ongoing work for https://outreach-io.atlassian.net/browse/DT-5272.

PR #712, opened 2026-07-09
feat(deployments): enable Istio Waypoint by default for gRPC services

AndrewWinterman:

Enables an Istio Waypoint proxy by default for all services with `grpc` in `serviceActivities`, outside of development environments.

Waypoints improve gRPC load balancing by spreading connections across all pods instead of pinning them to a subset. This prevents resource saturation (e.g. PostgreSQL connection pools) and improves HPA efficacy.

Rendered into `nonDevelopmentObjects` (staging/production only), mirroring the [`admingw.override.jsonnet`](https://github.com/getoutreach/admingw/blob/main/deployments/admingw/admingw.override.jsonnet#L141) pattern:

- Labels the k8s `Service` with `istio.io/use-waypoint: <app>-waypoint`
- Adds `WaypointProxy`, `WaypointProxyConfig`, and a `GrpcRoute` (backend port 5000)
- `team` sourced from the `reportingTeam` stencil arg

Adds a per-service opt-out via `kubernetes.istioWaypoint.enabled: false` in `service.yaml` (defaults to `true`), supporting a staggered rollout through Stencil on a per-k8s-service basis.

## Jira ID

[DT-5423](https://outreach-io.atlassian.net/browse/DT-5423)

## Notes for your reviewers

- Only the `TestRenderDeploymentJsonnet_Canary` snapshot changed — it's the only test case with `grpc` in `serviceActivities`. Non-gRPC render paths are unchanged (block is skipped).
- Objects only render outside development (`nonDevelopmentObjects`), matching the "outside of devenv" requirement.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DT-5423]: https://outreach-io.atlassian.net/browse/DT-5423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


manifest.yaml
```diff
@@ -97,6 +97,11 @@ arguments:
     schema:
       type: boolean
     description: "Only applicable when `service` is true. This is a flag to enable topology aware routing in K8s for your service."
+  kubernetes.istioWaypoint.enabled:
```
richardlarocque, line 100: Perhaps this should be `istioWaypointGrpc`?  Seems plausible that we'll be doing the same for HTTP or GraphQL endpoints at some point.

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
```diff
@@ -266,6 +266,34 @@ local all = {
...
+  // `kubernetes.istioWaypoint.enabled: false` in service.yaml.
+  service+: {
+    metadata+: {
+      labels+: {
+        'istio.io/use-waypoint': app.name + '-waypoint',
+      },
+    },
+  },
+  waypoint: ok.WaypointProxy(app.name + '-waypoint', namespace=app.namespace, team='test:team'),
+  waypointConfig: ok.WaypointProxyConfig(app.name + '-waypoint', namespace=app.namespace, team='test:team'),
```
richardlarocque, line 281: I was going to object to this on the ground that it would be nice to share waypoints.

My understanding of this config is that it will create envoy pods for every service to which it's applied.  That's kind of a waste at our typical service volumes.  I think we'd be better off with one big shared envoy as the default, and making dedicated envoys optional for cases that really need it.

I poked around in staging and it looks like this is the standard pattern, though.  I guess it's fine.  Would be nice to leverage the standardization to move away from it, though.

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
```diff
@@ -266,6 +266,34 @@ local all = {
...
+  grpcRoute: ok.GrpcRoute(app.name, app.namespace) {
```
richardlarocque, line 282: Is this necessary?  Does it do anything?

If not, it might be better to remove.  If someone did want to muck about with custom grpc routes, this thing would only get in the way.

richardlarocque (commented):
I noticed there's ~20 services in staging that are already using waypoints today.

This new stencil template might cause them some grief, because by default they'll get two waypoints.

I don't think that blocks this effort.  In the long term they should be happy that this stuff is baked in rather than extending their deployment specs.  Might want to give those teams a heads-up or post about it in `#dev-announce` as a fair warning, though.

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
AndrewWinterman: tbh sharing feels like more of a can of worms than i feel prepared for-- maybe it's already set up for it with the GrpcRoute custom resource ProxyConfig? but it seems like a can of worms

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
AndrewWinterman: I think so? honestly a bit unclear

manifest.yaml
AndrewWinterman: okay, did a spot of research:  

> proxies ALL service traffic, not gRPC-only. gRPC-specific part is the GrpcRoute + the has "grpc" gate, not the waypoint. So istioWaypointGrpc is wrong name for the thing. When HTTP arrives you likely reuse same waypoint, add HttpRoute. Generic name ages better. 

this also clarifies why you need the grpc route

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
AndrewWinterman: huh, https://istio.io/latest/docs/ambient/usage/waypoint/#usewaypointnamespace so it can do it but it'd be work for COR, and then we'd remove these and set istio.io/use-waypoint-namespace as well as istio.io/use-waypoint labels on the service. 

I'm not 100% certain what argo's behavior is removing objects, whether it leaves them present and unmanaged or deletes them, but depending we could switch to shared waypoints in a future restencil

manifest.yaml
richardlarocque: OK.  If this evolves into a general enable/disable of waypoint across gRPC, HTTP, GraphQL-over-HTTP, etc. then I would be OK with this name.

It may even be better that way, in the sense that if you want to do something weird and enable waypoints only for some of your exposed services, you can just turn this flag off and do your own config in the override manifests.  This moves the complexity into the service that's being weird and out of the stencil templates.

But the GrpcRoute is currently forcing it to be gRPC-only?  I guess the intent is to make it less restricted in the future?

richardlarocque (approved):
LGTM.  I'd be happy to ship some more waypoint support, even if it's not perfect yet.

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
jan-ludvik: I never considered cross namespace as way to sharing, I though it's only about using 1 waypoint for 1 service if we wanted to change namespace - for example for operations -> one team owns waypoints and provisions them for other teams but they cannot change its config directly.

templates/.snapshots/TestRenderDeploymentJsonnet_Canary-deployments-appname-app.jsonnet.tpl-deployments-testing-testing.jsonnet.snapshot
jan-ludvik: If I remember well we had an issue where we figured out this is needed otherwise traffic will be treated just as HTTP and not processed correctly

